PR #2, opened 2018-10-24
MES-936/932 - Prevent proceeding beyond declaration consent after clearing signature + remove recording consent

johnmorleybjss:
The enabled state of the continue button is determined by a combination of the insurance/residency checkboxes and the signature. Fixing failure to unset signature in component after clearing signature pad.


johnmorleybjss:
Forgot to create a different branch/PR for MES-932 (remove post-test debrief audio recording declaration). That ticket is also in this PR.

PR #10, opened 2018-10-29
Feature mes 1019 appium test framework

andrew-lai-bjss:


generateTestReport.js
```diff
@@ -0,0 +1,19 @@
...
+  reportSuiteAsScenarios: true,
+  launchReport: false
+  // metadata: {
+  //     "App Version":"0.3.2",
+  //     "Test Environment": "STAGING",
+  //     "Browser": "Chrome  54.0.2840.98",
+  //     "Platform": "Windows 10",
+  //     "Parallel": "Scenarios",
+  //     "Executed": "Remote"
+  // }
```
JamieKellyBJSS: How come this is all commented out?

test/e2e/step-definitions/generic-steps.ts
```diff
@@ -0,0 +1,77 @@
...
+  return expect(
+    element
+      .all(by.css('span.mes-header-md'))
+      .first()
+      .getText()
+  ).to.eventually.equal('Your Journal');
+});
+
+Given('I am logged in and on the journal page', function() {
+  var loginButton = element(by.xpath('//button[span[text()="Ok"]]'));
```
johnmorleybjss: Perhaps using IDs would be a better idea? That way if do a simple text change the tests don't break?

andrew-lai-bjss: If the intention is to rework all the elements with ID's then yes the use of ID's would be preferable

PR #13, opened 2018-10-31
Feature mes 955 test report screen new layout styling

inhand:
This PR basically skins the test form to the grey design from zeplin. It also forms the basis for the gds stuff. So defines colours etc.


inhand:
Screen dump of the new design
955-screen-dump

PR #14, opened 2018-11-01
MES-941: Prevent proceed beyond office screen before mandatory fields populated

johnmorleybjss:
* Ensure that route number, weather, show me question, independent driving and physical description are all completed.
* In the event of a fail due to dangerous/serious, ensure that each of those faults has an associated note.
* ~~Disable the submit button unless the above criteria are met.~~
* Give an alert when submit is pressed without the form being complete. Ensure the examiner _can_ defer completing the form if they have another imminent test.

### DIFF
--- a/src/components/post-test-summary-section/post-test-summary-section.ts
+++ b/src/components/post-test-summary-section/post-test-summary-section.ts
@@ -2,9 +2,10 @@ import { IManualSummary } from './../test-summary/interfaces/IManualSummary';
 import { Component, Input } from '@angular/core';
 import { IFaultSummary } from '../test-summary/interfaces/IFaultSummary';
 import { FaultTitle } from '../test-summary/enums/FaultTitle';
-import { find } from 'lodash';
+import { find, isNil } from 'lodash';
 import { ModalController } from 'ionic-angular';
 import { TextboxModalComponent } from '../textbox-modal/textbox-modal';
+import { isNonBlankString } from '../../shared/utils/string-utils';
 
 @Component({
   selector: 'post-test-summary-section',
@@ -13,6 +14,7 @@ import { TextboxModalComponent } from '../textbox-modal/textbox-modal';
 export class PostTestSummarySectionComponent {
   @Input() summary: IFaultSummary;
   @Input() manualSummary: IManualSummary;
+  @Input() canComplete: boolean = false;
   drivingFaultsTitle: string = FaultTitle.DrivingFaults;
   // Map of fault name to the note
   private faultNotes = new Map<string, string>();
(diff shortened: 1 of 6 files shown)


src/components/post-test-summary-section/post-test-summary-section.ts
```diff
@@ -37,4 +39,16 @@ export class PostTestSummarySectionComponent {
     textboxModal.onDidDismiss((notes?: string) => (this.faultNotes[faultName] = notes));
     textboxModal.present();
   }
+
+  isComplete(): boolean {
+    if (!this.canComplete) {
+      return true;
+    }
+    const notes = Object.keys(this.faultNotes)
```
ammarhaiderbjss: Could use `Object.values` instead? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values

src/pages/post-test-summary/post-test-summary.ts
```diff
@@ -72,7 +79,33 @@ export class PostTestSummaryPage {
...
+        .create({
+          title,
+          message,
+          buttons: [
+            {
+              text: 'Cancel'
+            },
+            {
+              text: 'Defer',
+              handler: () => this.backToJournal()
```
ammarhaiderbjss: Do we need this additional wrapper? can we just pass `this.backToJournal` instead?

src/components/post-test-summary-section/post-test-summary-section.ts
johnmorleybjss: Attempted to use this first off, but we don't have support, we'd need to configure TS for ES2017. Happy to make the change if that's what we want.

src/pages/post-test-summary/post-test-summary.ts
johnmorleybjss: I tried without the wrapping arrow, I had a `this` binding issue.

src/components/post-test-summary-section/post-test-summary-section.ts
ammarhaiderbjss: ok nvm

src/pages/post-test-summary/post-test-summary.ts
ammarhaiderbjss: fair enough, missed that this was all happening in a`.create`

src/components/post-test-summary-section/post-test-summary-section.ts
johnmorleybjss: Upgraded to ES2017, confirmed that `Object.values` is working as expected.

PR #19, opened 2018-11-05
Moves S&D buttons and associated functionality into report header component

ammarhaiderbjss:
image


Cavalier07:
How about the replacement location for the name - or is that covered elsewhere?

PR #24, opened 2018-11-06
MES-1042: Waiting room GDS rework

johnmorleybjss:
* Remove alpha banner and set fixed height gray top navigation bar
* Change primary app background colour to gray
* Extend waiting room card to include name and driver number
* Add additional end driving test secondary CTA
* Restyle primary CTA to fit GDS guidelines
* Refactor signature pad into its own component which includes the label and retry button
* Style waiting room screen as per designs

img_9dd7235d9094-1


src/pages/all-on-one-v2/all-on-one-v2.ts
```diff
@@ -143,7 +151,7 @@ export class AllOnOneV2Page {
    */
   getTitle(): string {
     if (this.slotDetail) {
-      return getFormattedCandidateName(this.slotDetail.candidateName) + ' - Test Report';
+      return `${getFormattedCandidateName(this.slotDetail.candidateName)} - Test Report`;
```
szabolcs-keresztes: Like this!

src/components/mes-signature-pad/mes-signature-pad.scss
```diff
@@ -0,0 +1,25 @@
+mes-signature-pad {
+  button.mes-signature-pad-retry-button {
+    font-weight: bold;
+    font-size: 17px;
+    border-radius: 10px;
+    float: right;
+    position: relative;
+    top: 3.5em;
```
szabolcs-keresztes: Was wondering if we could use only one measuring unit, `px` or `em`?

johnmorleybjss: You make a good point. I've been trying to use `em` where possible, but in this case, I didn't think I'd want the border radius to adjust along with the font size. Probably worth a wider discussion with the team. 👍

ammarhaiderbjss:
Can we get a screenshot from app on the description please?

johnmorleybjss:
Adding a screenshot.
A few notes:
* @rstepnowski expressed an interest in picking up the last few changes on this ticket, these include:
    * Fixing the fonts
    * Implementing the GDS checkboxes
    * Fixing any spacing issues
* However, after showing the POs yesterday, they say this is "close enough" for familiarisation. We all agreed that we need to decide whether we make implementing all the necessary GDS components a technical requirement even if the POs are happy without them.

PR #26, opened 2018-11-07
MES-1036: UI rework - CONTACT DETAILS

rstepnowski:


src/pages/candidate-info/candidate-info.html
```diff
@@ -1,86 +1,119 @@
...
+                    Category B
+                </ion-col>
+            </ion-row>
+            <div class="mes-divider"></div>
+            <ion-row>
+                <ion-col col-4>
+                    Slot
+                </ion-col>
+                <ion-col col-8>
+                    {{getSLotType()}}
```
szabolcs-keresztes: getSlotType - might be good to write a lowercase L here.

src/pages/candidate-info/candidate-info.html
```diff
@@ -1,86 +1,119 @@
...
+                </ion-col>
+            </ion-row>
+            <div class="mes-divider"></div>
+            <ion-row>
+                <ion-col col-4>
+                    Candidate comments
+                </ion-col>
+                <ion-col col-8>
+                    <span>
+                        <img src="assets/imgs/warning.png" class="warning-icon" *ngIf="showSlotWarning(s)" />
```
szabolcs-keresztes: What would the `s` be in the `showSlotWarning` function parameter?

src/shared/utils/formatters.ts
```diff
@@ -9,16 +9,15 @@ export function getFormattedCandidateName(candidateName: ICandidateName): string
 
 /**
  * `Builds a formatted html structure from a given IAddress
- * todo - this is quick and dirty for now. Needs to deal with missing fields etc
```
ammarhaiderbjss: Can we keep the comment that this is quick and dirty? We should really avoid having markup in .ts files. HTML belongs in .html files. Think this was something done in haste and should be discouraged.

src/pages/candidate-info/candidate-info.html
rstepnowski: Unused parameter - just realised it :) It's supposed to be the slotDetails object but the function actually class property ;) I'll remove it
Nice catch - thanks

src/shared/utils/formatters.ts
rstepnowski: ok -done

src/pages/candidate-info/candidate-info.html
rstepnowski: fixed 👍

src/pages/candidate-info/candidate-info.html
```diff
@@ -1,86 +1,119 @@
...
+
     <div class="confirm-identity-box" *ngIf="signature">
         <div class="inner" (click)="confirmIdentity()">
             <ion-checkbox></ion-checkbox>
             <button class="mes-button-primary">Confirm Candidates Identity</button>
         </div>
     </div>
+
+    <div class="mes-view-candidate-header-padding-fix">
+        <button float-left class="mes-cta-button-secondary" outline color="dark">
```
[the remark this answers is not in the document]
rstepnowski: Yes, you're correct - I've removed that, thanks 👍

PR #34, opened 2018-11-12
github pull request template

rstepnowski:
A proposal of Github pull request template

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,25 @@
+## Description and relevant Jira numbers
+
+## Pull Request checklist:
+
+- [ ] [WIP] tag removed from PR title
+- [ ] PR has an understandable description
+
+## Git feature branch checklist:
+
+- [ ] branch name comply with our branching strategy
+- [ ] git branch contains relevant JIRA ticket number
+- [ ] branch rebased against the latest develop
+- [ ] commits are squashed
+
+## Sign off process checklist:
+
+- [ ] Code has been tested manually
+- [ ] Tests are passing
+- [ ] PR link added to JIRA
+- [ ] Reviewers selected in Github
+- [ ] Any new reusable component/widget added to component library on Confluence
+
+- [ ] Screenshot(s) captured on the physical device (if applicable)
+- [ ] Tested by QA
+- [ ] PO's approval


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Sign off process checklist:
+
+- [ ] Code been tested manually
+- [ ] Tests are passing
+- [ ] PR link added to JIRA
+- [ ] Tests are passing
+- [ ] Reviewers selected in Github
+
+- [ ] Screenshot (if applicable)
```
ammarhaiderbjss: can we specify this needs to be from the physical device?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Pull Request checklist:
+
+- [ ] [WIP] tag removed from PR title
+- [ ] PR has an understandable description
+
+## Git feature branch checklist:
+
+- [ ] branch rebased against the latest develop
+- [ ] commits contain relevant JIRA ticket number
```
ammarhaiderbjss: Commits or branch name?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,24 @@
...
+- [ ] commits are squashed
+
+## Sign off process checklist:
+
+- [ ] Code been tested manually
+- [ ] Tests are passing
+- [ ] PR link added to JIRA
+- [ ] Tests are passing
```
ammarhaiderbjss: Can get rid of one 'Tests are passing'

.github/PULL_REQUEST_TEMPLATE.md
rstepnowski: I've meant commit messages here but i can add another point for branch naming strategy since this also applicable

.github/PULL_REQUEST_TEMPLATE.md
rstepnowski: done

.github/PULL_REQUEST_TEMPLATE.md
rstepnowski: done

.github/PULL_REQUEST_TEMPLATE.md
szabolcs-keresztes: Is it necessary for the commit message to contain the ticket number?

ammarhaiderbjss: yeah, feels like it could become a pain

Cavalier07: If we are squashing commits there should only be a couple of commits per branch at max

rstepnowski: It's a matter o habit imho - additional 6-7 chars in commit msg won't kill anyone i guess :}
Especially if we squash commits at the end - amending at least one of them should not be a hassle.

So... do we want to keep it or drop it?

johnmorleybjss: I don't want to put the ticket number on every commit. The fact that the ticket number is included in the branch name should be enough. If we're using GitHub's built-in squash merge, I don't think we can't enforce a squash commit message before merging anyway.

szabolcs-keresztes: If we really want to hold on to the ticket number we could create a **tag** whenever the specific story has been implemented.

Just an idea, but it feels like it would complicate things with only small benefits in return.

rstepnowski: Branches come and go (eg. get deleted) but commits stay in git log "forever"  ;)

The rationale behind Jira ticket number in commit msg is to improve traceability of when given line was modified and why. It links every line change with a story in Jira and therefore provides a bit of more context info when you stare at some particular line of code and try to answer the question "why the author did it the way he did".

As long as we squash all commits on a feature branch and at least one of them contains Jira ticket that would suffice imho. 

I just would like to avoid small commits with short/undescriptive messages that don't link with anything...
I won't force it if the majority are against it - I just consider it as a quality of life improvement for long-lasting projects

szabolcs-keresztes: We could link the jira ticket to the PR?

johnmorleybjss: I agree with your sentiment @rstepnowski, when looking at `develop` branch history, it's nice to see a clean timeline of ticket completion. However, when working on a feature branch, sometimes I like to do _lots_ of small commits. As long as we squash the feature branch when it's merged and that squash commit includes the ticket number, I think we're good.

szabolcs-keresztes: I think people expressed their opinions around the issue. And we as a team came up with great solutions. 
So let's make a decision, keep to it in the future and complete this PR.

rstepnowski: It was agreed on slack chat that we will skip this one - I'll remove this one and merge it.

PR #49, opened 2018-11-16
MES-1052 - Help Guide List

szabolcs-keresztes:

Implementing the Help Guide list. Which hast 5 sections and all of them are expandable and they contain additional information which is a generic youtube video at the moment. Because the content is not ready yet.

img_0042

## Pull Request checklist:

- [x] commits are squashed

- [x] Tests are passing
- [x] PR link added to JIRA

- [x] Tested by QA
- [x] PO's approval

### DIFF
--- a/src/help/pages/help-section/help-section.scss
+++ b/src/help/pages/help-section/help-section.scss
@@ -1,2 +1,7 @@
 page-help-section {
+  .show-icon,
+  .hide-icon {
+    font-size: 24px;
+    font-weight: bold;
+  }
 }
--- a/src/help/pages/help-section/help-section.scss
+++ b/src/help/pages/help-section/help-section.scss
@@ -5,6 +5,14 @@ page-help-section {
     font-weight: bold;
   }
 
+  .section-number {
+    font-size: 36px;
+    font-weight: bold;
+    border: 3px solid black;
+    border-radius: 50%;
+    padding: 0px 11px;
+  }
+
   #show-all-container {
     width: 100%;
     text-align: right;
@@ -12,5 +20,29 @@ page-help-section {
 
   .hide-show-links {
     float: right;
+  }
+
+  .divider {
+    @extend .mes-divider;
+
+    margin-left: 80px;
+  }
+
+  .section-header-text {
+    @extend .mes-header-lg;
+
+    margin-left: 28px;
+  }
+
+  .section-header {
+    margin-bottom: 15px;
+  }
+
+  .section-text {
+    margin-left: 80px;
+  }
+
+  .section-details {
+    margin-left: 80px;
   }
 }
--- a/src/help/pages/help-section/help-section.scss
+++ b/src/help/pages/help-section/help-section.scss
@@ -4,4 +4,9 @@ page-help-section {
     font-size: 24px;
     font-weight: bold;
   }
+
+  #show-all-container {
+    width: 100%;
+    text-align: right;
+  }
 }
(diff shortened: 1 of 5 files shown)


src/help/pages/help-section/help-section.scss
```diff
@@ -1,2 +1,126 @@
...
+
+  .header-container {
+    padding-bottom: 38px;
+  }
+
+  #help-guide-content {
+    padding: 16px;
+  }
+
+  .section-item::after {
```
johnmorleybjss: Just looked at the designs, can you do an `&:last-child` to omit the line after the final section?

src/help/pages/help-section/help-section.ts
```diff
@@ -1,23 +1,235 @@
...
+  // Expanding and collapsing Journal section
+
+  isJournalVisibilityExpended() {
+    return this.visibility.journal === visibilityType.expanded;
+  }
+
+  isJournalVisibilityCollapsed() {
+    return this.visibility.journal === visibilityType.collapsed;
+  }
+
```
Cavalier07: Why do we need 2 methods for a toggle with only 2 options shouldn't we be just  isJournalVisibilityExpended() and !isJournalVisibilityExpended() to show and hide.

To me this seems to be over engineered where a boolean for each toggle would have been enough.

src/help/pages/help-section/help-section.ts
```diff
@@ -1,23 +1,235 @@
...
+  toggleJournalVisibility() {
+    if (this.isJournalVisibilityCollapsed()) {
+      this.visibility.journal = visibilityType.expanded;
+      return;
+    }
+    if (this.isJournalVisibilityExpended()) {
```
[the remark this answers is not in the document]
szabolcs-keresztes: Changed it.

src/help/pages/help-section/help-section.scss
johnmorleybjss: Just had a go myself, doesn't look quite as easy as I made out...

src/help/pages/help-section/help-section.ts
szabolcs-keresztes: Yes, you are right that for this specific example (if it never going to change) using a boolean would have been a solution. However I didn't choose that solution because in the majority of the cases when you start toggling boolean to represent the state of your application with that type is going to end up having bugs. For instant you, as a reader don't know that `false` means expanded or collapsed, but with enum is clear.

So this is a different way of thinking which is helping me to have a pure business logic around this list.

src/help/pages/help-section/help-section.scss
szabolcs-keresztes: I will put an ending on the line instead of removing it.

src/help/pages/help-section/help-section.ts
```diff
@@ -1,23 +1,235 @@
...
+  isWaitingRoomToCarVisibilityExpended() {
+    return this.visibility.waitingRoomToCar === visibilityType.expanded;
+  }
+
+  isWaitingRoomToCarVisibilityCollapsed() {
+    return this.visibility.waitingRoomToCar === visibilityType.collapsed;
+  }
+
+  toggleWaitingRoomToCarVisibility() {
+    if (this.isWaitingRoomToCarVisibilityCollapsed()) {
```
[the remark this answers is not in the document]
szabolcs-keresztes: Cool idea, will introduce a utility function.

src/help/pages/help-section/help-section.ts
johnmorleybjss: I like the use of enums to enrich primitive values with more meaning and to increase readability. I think we should start using this style more often.

PR #57, opened 2018-11-21
Fix scrolling of journal page on long dates

johnmorleybjss:
On long days, there was wrapping occurring on the date at the top of the journal page, causing the page to scroll. Increase the size the centre column to prevent this.


johnmorleybjss:
Tested with Wendesday 28th September (the longest date string we could have) and the wrapping is fixed.

PR #87, opened 2018-12-12
MES-1260 - Added Authentication

Cavalier07:

Adds authentication to the app.

## Pull Request checklist:

- [ ] [WIP] tag removed from PR title

- [ ] commits are squashed

- [x] Tests are passing
- [x] PR link added to JIRA

- [ ] Tested by QA
- [x] PO's approval


src/classes/base-page.ts
```diff
@@ -0,0 +1,34 @@
+import { Platform, NavController } from 'ionic-angular';
+import { AuthenticationProvider } from '../providers/authentication/authentication';
+
+export abstract class BasePageComponent {
```
ammarhaiderbjss: do we know what this keyword does?

src/providers/authentication/authentication.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  constructor(
+    private msAdal: MSAdal,
+    private inAppBrowser: InAppBrowser,
+    appConfig: AppConfigProvider) {
+    this.authenticationSettings = appConfig.getAppConfig().authentication;
+  }
+
+  isAuthenticated = (): boolean => {
+    return this.authenticationToken ? true : false;
```
ammarhaiderbjss: Need to have a think about how to make this a better check for whether a user is logged in or not, should have a look in the adal library's source code to see if there's anything we can use there. Failing that, we could decode the token and check for some essential values.

src/classes/base-page.ts
Cavalier07: abstract ? - Basically we are unable to create an instance of it, the only way to use it is to create a sub class (All pages will extend this class). We are basically using it give the same functionality to multiple pages and if needed in the future can use it to force pages to implement certain methods.

src/providers/authentication/authentication.ts
Cavalier07: Agreed - This is something we need to investigate - could it be worth spinning this off into a second story - Azure AD Token Validation in Mobile App . I'm wary of pushing more into this  ticket

src/pages/login/__tests__/login.spec.ts
```diff
@@ -34,6 +37,11 @@ describe('LoginPage', () => {
     it('should create', () => {
       expect(component).toBeDefined();
     });
+    it('should login successfully', () => {
+      component.login().then(() => {
+        expect(navController.setRoot).toBeCalled;
```
johnmorleybjss: The call seems to be missing here, I don't think you're checking anything. Would be nice to test it's called with the correct argument too.

src/providers/app-config/app-config.ts
```diff
@@ -34,6 +34,13 @@ export class AppConfigProvider {
       this.appConfig = {
```
johnmorleybjss: Perhaps we could do something about this config call failing? I imagine it's quite a common scenario (e.g. when offline).

src/providers/authentication/__mocks__/ms-adal.mock.ts
```diff
@@ -0,0 +1,42 @@
+// TODO: Move it to providers/authentication/__mocks__/
```
johnmorleybjss: Think you've addressed this TODO.

src/providers/authentication/__tests__/authentication.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+import { AuthenticationProvider } from '../authentication';
+import { AppConfigProvider } from '../../app-config/app-config';
+import { AppConfigProviderMock } from '../../app-config/__mocks__/app-config.mock';
+import { InAppBrowserMock } from '../__mocks__/in-app-browser.mock';
+
+describe('Authentication', () => {
+  let authenticationProvider: AuthenticationProvider;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
```
johnmorleybjss: Do you need TestBed to test this class? Or is it just easier than managing the mock lifecycle by hand?

src/providers/authentication/__tests__/authentication.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    it('should sign in with credetials', async() => {
+      await authenticationProvider.loginWithCredentials();
+
+      expect(authenticationProvider.isAuthenticated()).toEqual(true);
+      expect(authenticationProvider.getAuthenticationToken()).toEqual('AYSNC TEST TOKEN');
+    });
+
+    it('should do something when login fails', async () => {
+      // TODO
```
johnmorleybjss: Seems quite an important test to elaborate.

Cavalier07: The function in the authentication provide it doesn't do anything either, I guess we can use it for logging or retries if we decide to implement them. It didn't seem right to restrict access/redirect people from here (I think it's better the pages control that) - The other option was to remove that test but I think it works as a reminder this is something we need to look at in the future

package.json
```diff
@@ -43,19 +43,25 @@
...
     "cordova-plugin-device": "^2.0.2",
+    "cordova-plugin-inappbrowser": "3.0.0",
     "cordova-plugin-ionic-keyboard": "^2.1.3",
     "cordova-plugin-ionic-webview": "^2.2.5",
+    "cordova-plugin-ms-adal": "0.10.1",
     "cordova-plugin-splashscreen": "^5.0.2",
     "cordova-plugin-statusbar": "^2.4.2",
     "cordova-plugin-whitelist": "^1.3.3",
     "event-stream": "3.3.4",
+    "io": "1.0.0",
```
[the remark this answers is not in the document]
Cavalier07: My guess is it snuck in have removed it and it didn't seem to break anything

src/pages/login/__tests__/login.spec.ts
Cavalier07: Fixed

PR #114, opened 2019-01-15
MES-986: Display last sync time on journal

johnmorleybjss:
MES-986: Display last sync time on journal

We already had this timestamp stored in the state for the purposes of populating the `If-Modified-Since` header on refresh requests. Simply extending the page state and template to display it.

Some additional cosmetic tweaks:
* Change status bar to black with white foreground.
* Move Refresh button into navbar.
* Move logout button to left side of the navbar, append the employee ID.
* Remove Waiting Room shortcut and use the "Start test" button to begin the dummy test journey instead.

img_943981acdfd9-1

## Pull Request checklist:

- [x] commits are squashed

- [x] Tests are passing
- [x] PR link added to JIRA

- [x] Tested by QA
- [x] PO's approval


johnmorleybjss:
UX seem to want the logout button moving into a hamburger menu on the left of the navbar, however, there's nothing in the ACs for this story about implementing this. Do people think we should tackle this as part of this story? If so, perhaps we could merge as partial completion as per suggested improvements in retro.

src/app/app.component.ts
```diff
@@ -20,8 +20,9 @@ export class App {
     private authenticationProvider: AuthenticationProvider,
   ) {
     platform.ready().then(() => {
-      statusBar.styleDefault();
+      statusBar.styleLightContent();
       statusBar.overlaysWebView(false);
+      statusBar.backgroundColorByHexString('#000000');
```
szabolcs-keresztes: Should this be part of the SCSS, and maybe using the existing GDS colours?

src/pages/journal/journal.scss
```diff
@@ -2,4 +2,14 @@ page-journal {
   h3 {
     text-align: center;
   }
+  .last-sync {
+    color: #dbd8d8;
```
szabolcs-keresztes: Can we use the GDS colours array for setting the this variable please?

src/pages/journal/journal.ts
```diff
@@ -7,19 +7,22 @@ import { BasePageComponent } from '../../classes/base-page';
...
 import { merge } from 'rxjs/observable/merge';
 import { SlotItem } from '../../providers/slot-selector/slot-item';
+import { isNil } from 'lodash';
+import * as moment from 'moment';
 
 interface JournalPageState {
   testSlots$: Observable<SlotItem[]>,
   error$: Observable<MesError>,
   isLoading$: Observable<boolean>,
+  lastRefreshed$: Observable<Date>,
```
szabolcs-keresztes: Would be more accurate to call this `lastRefreshedTime$` so that we know that it refers to the actual time.

src/pages/journal/journal.ts
```diff
@@ -157,6 +164,13 @@ export class JournalPage extends BasePageComponent implements OnInit, OnDestroy
     this.loadJournal();
   };
 
+  public formatLastRefreshTime() {
```
szabolcs-keresztes: You could pipe this function to the outcome of `getLastRefreshed` (which would be best to call `getLastRefreshedTime`) so that we don't need to call this function in the html template but just get the lastRefreshedTime$ asynchronously.

What do you think?

szabolcs-keresztes:
I think the burger menu can come as a separate piece of work.

src/app/app.component.ts
johnmorleybjss: I spent a bit of time looking for a solution in SCSS but couldn't find one, probably because this interface is through a Cordova plugin, not Ionic itself. I'll have a bit more of a read around for a better solution.

johnmorleybjss:
Yep @szabolcs-keresztes, agreed on the JIRA ticket that we're not touching the menu on this ticket.

src/pages/journal/journal.scss
johnmorleybjss: I'm not sure this is a GDS colour, as I don't see this value in `$colors-gds`. That said, I don't know where the values for `$colors-gds` came from, it could be that some standard colours are missing.

I considered adding to `$colors`, but didn't bother in the end as I didn't foresee the colour being reused - are we mandating that _all_ colours are added as SASS variables? Personally, I'd prefer the approach of using colour codes directly and doing a quick search through the codebase for existing usages whenever you introduce a colour (at which point you can extract a variable). Thoughts?

src/pages/journal/journal.ts
johnmorleybjss: Done.

src/pages/journal/journal.scss
szabolcs-keresztes: I think all colours should be reused, and we should have a colour palette. There is no need for colours being used only in one place.

But let's leave this as it is right now, and have a discussion with Nad and Oli. They will be able to talk us through the colour palette. And by doing that I'm pretty sure we can remove quite a few of the already registered colours in `$colors-gds`.

src/pages/journal/journal.ts
johnmorleybjss: Are you referring to storing strictly the formatted date string in the page state? While this would make the template cleaner, I think that a `Date` type in the page state is a better solution as this prevents dataloss being propagated into the state. The reason being, we have a story in the upcoming sprint to start using the last refresh datetime as a basis for automatic refresh - this would be tougher if we store a formatted string in the state.

johnmorleybjss: If I have misinterpreted the suggested solution, please let me know!

src/pages/journal/journal.scss
szabolcs-keresztes: Also I'm not sure if new colours should be introduced just like that outside of that colour palette that we are going to establish with the designers.

src/app/app.component.ts
szabolcs-keresztes: I see, that's fine then

src/pages/journal/journal.ts
johnmorleybjss: After discussion, the proposed solution _is_ to simply keep a string of the formatted date in the journal page state. The reason for this is down to the intended usage of the page state, which should, wherever possible, be purely representational. I will refactor to adhere to this principle.

PR #138, opened 2019-01-30
MES-1584: Begin migration From Jest to Jasmine

johnmorleybjss:
It has been decided by the powers that be that Jest is forbidden and our unit tests _must_ be in Jasmine.

Begin the process of migration by running both Jest and Jasmine + Karma until all tests are converted over.

Pretty sure this will need some rework in order to play nicely with our automated pipeline, as it uses Karma to launch Chrome for the Jasmine tests.

* Rename all existing tests to have the suffix `.spec.jest.ts`, reconfigure Jest to run them.
* Add dependencies and config for Jasmine + Karma as specified [here](https://leifwells.github.io/2017/08/27/testing-in-ionic-configure-existing-projects-for-testing/).
  * This guide is written for both unit and e2e tests, so there may be some overlap with existing e2e tests that can be factored out.
  * The config has been copied verbatim, much of it may not be needed.
* Convert a couple of test suits back to `.spec.ts`, having them run via Jasmine.
* Update `npm test` script to run both test frameworks.


johnmorleybjss:
At some point, we should switch to using [jasmine-spec-reporter](https://www.npmjs.com/package/jasmine-spec-reporter). We could also potentially look at [karma-jsdom-launcher](https://www.npmjs.com/package/karma-jsdom-launcher) to avoid having to launch Chrome.

PR #142, opened 2019-02-01
Implementing the navigation through the weekends

szabolcs-keresztes:

Implementing the functionality of inserting empty days and also the navigation logic when it's a simple weekday or it's a Friday.

## Pull Request checklist:

- [x] commits are squashed

- [x] Tests are passing
- [x] PR link added to JIRA

- [x] Tested by QA
- [x] PO's approval

### DIFF
--- a/src/pages/journal/journal.effects.ts
+++ b/src/pages/journal/journal.effects.ts
@@ -45,6 +45,8 @@ export class JournalEffects {
         .pipe(
           map((journalData: ExaminerWorkSchedule) => this.slotProvider.detectSlotChanges(slots, journalData)),
           map((slots: any[]) => groupBy(slots, this.slotProvider.getSlotDate)),
+          map((slots: {[k: string]: SlotItem[]}) => this.slotProvider.extendWithEmptyDays(slots)),
+          map((slots: {[k: string]: SlotItem[]}) => this.slotProvider.getRelevantDays(slots)),
           map((slots: {[k: string]: SlotItem[]}) => new journalActions.LoadJournalSuccess(slots)),
           catchError(err => of(new journalActions.LoadJournalFailure(err)))
         );
--- a/src/pages/journal/journal.effects.ts
+++ b/src/pages/journal/journal.effects.ts
@@ -46,7 +46,7 @@ export class JournalEffects {
           map((journalData: ExaminerWorkSchedule) => this.slotProvider.detectSlotChanges(slots, journalData)),
           map((slots: any[]) => groupBy(slots, this.slotProvider.getSlotDate)),
           map((slots: {[k: string]: SlotItem[]}) => this.slotProvider.extendWithEmptyDays(slots)),
-          map((slots: {[k: string]: SlotItem[]}) => this.slotProvider.getRelevantDays(slots)),
+          map((slots: {[k: string]: SlotItem[]}) => this.slotProvider.getRelevantSlots(slots)),
           map((slots: {[k: string]: SlotItem[]}) => new journalActions.LoadJournalSuccess(slots)),
           catchError(err => of(new journalActions.LoadJournalFailure(err)))
         );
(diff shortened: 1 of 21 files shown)


src/pages/journal/journal.ts
```diff
@@ -119,11 +119,13 @@ export class JournalPage extends BasePageComponent implements OnInit, OnDestroy
   };
 
   private createSlots = (emission: any) => {
-    if (!Array.isArray(emission) || emission.length === 0) {
-      return;
-    }
+    if (!Array.isArray(emission)) return;
```
johnmorleybjss: I like this change to the subscription conceptually. Personally not a fan of inlining the return, but I suppose the style standard is one for the wider team.

src/pages/journal/journal.effects.ts
```diff
@@ -45,6 +45,8 @@ export class JournalEffects {
         .pipe(
           map((journalData: ExaminerWorkSchedule) => this.slotProvider.detectSlotChanges(slots, journalData)),
           map((slots: any[]) => groupBy(slots, this.slotProvider.getSlotDate)),
+          map((slots: {[k: string]: SlotItem[]}) => this.slotProvider.extendWithEmptyDays(slots)),
+          map((slots: {[k: string]: SlotItem[]}) => this.slotProvider.getRelevantSlots(slots)),
```
johnmorleybjss: Can you foresee using these two methods of the slot provider independently, or are they completely coupled? If they are coupled, perhaps you could refactor to only have to do a single mapping here?

src/providers/slot/slot.ts
```diff
@@ -39,6 +39,47 @@ export class SlotProvider {
     });
   }
 
+  /**
+   * Exentind the journal with empty days where there was no slots defined in the next 7 days
```
ammarhaiderbjss: "Exentind"? lol

src/pages/journal/journal.ts
ammarhaiderbjss: I like inline returns, its my guilty JS pleasure

src/providers/slot/slot.ts
```diff
@@ -39,6 +39,47 @@ export class SlotProvider {
     });
   }
 
+  /**
+   * Extending the journal with empty days where there was no slots defined in the next 7 days
+   * @param slots Journal slots
+   * @returns Slots with additional empty days 
+   */
+  extendWithEmptyDays = (slots: {[k: string]: SlotItem[]}): {[k: string]: SlotItem[]} => {
+    const emptyDays = {
```
ammarhaiderbjss: Think the intention is to have this 7 day value configurable. In which case we should have a function that reads from config and generates these empty days.

src/providers/slot/slot.ts
```diff
@@ -39,6 +39,47 @@ export class SlotProvider {
...
+  }
+
+  /**
+   * Slice the journal slots and get the slots only for the relevant days
+   * | From regular working weekday we can see the next working weekday
+   * | From Friday we can navigate through weekend till Monday
+   * @param slots Journal slots
+   * @returns Only the relevant slots
+   */
+  getRelevantSlots = (slots: {[k: string]: SlotItem[]}): {[k: string]: SlotItem[]} => {
```
ammarhaiderbjss: do we have unit tests for these new methods?

src/providers/slot/slot.ts
szabolcs-keresztes: Good point I moved to the value to environment variables which will be provided by the app config provider. But it's not a very good pace for it I think. Any other idea?

src/providers/slot/slot.ts
szabolcs-keresztes: We have one test for `extendWithEmptyDays` function. But we don't have any for the `getRelevantSlots` and the reason for this is that function is heavily dependent on `moment()`. So by this we are highlighting the fact that we need to abstract the way that we are working with date and time and can't just have our code being polluted by `moment.js`.

szabolcs-keresztes: We are going to build an abstraction layer around `moment.js` so that we can control the features better. And that time will write the tests for all the different cases what we left out while we have been depending on the framework.

PR #174, opened 2019-03-01
Mes 1721 send logs

szabolcs-keresztes:

Sending logs every minute.

## Pull Request checklist:

- [x] commits are squashed

- [x] Tests are passing
- [x] PR link added to JIRA

- [ ] Tested by QA
- [ ] PO's approval


src/modules/logs/logs.actions.ts
```diff
@@ -1,12 +1,38 @@
 import { Action } from '@ngrx/store';
 import { Log } from '../../shared/models/log.model';
 
-export const SAVE_LOG = 'Save Log';
+export const SAVE_LOG = '[Anywhere] Save Log';
```
Cavalier07: How about [GLOBAL] or [SHARED]? I don't really like Anywhere

src/modules/logs/logs.effects.ts
```diff
@@ -0,0 +1,58 @@
+import { Injectable } from '@angular/core';
+import { Effect, Actions, ofType } from '@ngrx/effects';
+
+import { switchMap, map, catchError, mapTo, withLatestFrom } from 'rxjs/operators';
+import { of } from 'rxjs/observable/of';
+import { interval } from 'rxjs/observable/interval';
+import { Store, select } from '@ngrx/store';
```
Cavalier07: Could this import be moved to the top group of imports.

src/providers/url/url.ts
```diff
@@ -15,7 +15,7 @@ export class UrlProvider {
   }
 
   getLoggingServiceUrl(): any {
-    return this.appConfigProvider.getAppConfig().loggingUrl;
+    return this.appConfigProvider.getAppConfig().logging.url;
```
Cavalier07: What are your thoughts about changing this to getLogsServiceUrl . It seems like everything else has moved from logging to logs

src/modules/logs/logs.effects.ts
```diff
@@ -0,0 +1,58 @@
...
+  @Effect()
+  sendLogsEffect$ = this.actions$.pipe(
+    ofType(logsActions.SEND_LOGS),
+    withLatestFrom(
+      this.store$.pipe(
+        select(getLogsState),
+      ),
+    ),
+    switchMap(([action, logs]) => {
+      console.log(logs);
```
Cavalier07: Console Log

src/providers/logs/logs.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Log } from '../../shared/models/log.model';
+import { UrlProvider } from '../url/url';
+import { Observable } from 'rxjs/Observable';
+
+@Injectable()
+export class LogsProvider {
+
+  constructor(public http: HttpClient, public urlProvider: UrlProvider) {}
+
+  public log = (log: Log): Observable<Object> => {
```
Cavalier07: This function is only used in the test now as the effect calls logMultiple() - is there a use where we would only send one log to the service , bypassing the store.

src/providers/logs/logs.ts
```diff
@@ -0,0 +1,20 @@
...
+    return this.logMultiple([log]);
+  }
+
+  public logMultiple = (logs : Log[]): Observable<Object> => {
+    const loggingServiceUrl = this.urlProvider.getLoggingServiceUrl();
```
Cavalier07: Is there a better return type you can use instead of Object

johnmorleybjss:
Should there be code in a reducer that cleans out the logs once they have been submitted?

szabolcs-keresztes:
@johnmorleybjss yes, there is going to be functionality in place which deletes the already sent logs, coming as part of a next PR.

src/modules/logs/logs.actions.ts
szabolcs-keresztes: We can agree on *[GLOBAL]* for these kind of actions, I like it.

src/modules/logs/logs.effects.ts
szabolcs-keresztes: sure

src/providers/url/url.ts
szabolcs-keresztes: Also the microservice is called `mes-logs-service` so ye it totally makes sense. But that would require an additional PR in the configuration microservice. Do you think that it's worth it?

src/providers/logs/logs.ts
szabolcs-keresztes: I think we are going to have the SEND_LOGS action sending logs for us. And that works with multiple even if there is one log in the array. So I'll delete the `log` function. And I also going to rename `logMultiple` to `sendLogs` function.

src/providers/url/url.ts
szabolcs-keresztes: Changed it and refactored the code so that we use `logs` everywhere (Global hah).

PR #181, opened 2019-03-06
Feature mes 1608 block app usage on unsupported devices

rossfauldsbjss:
MES 1608 - block app usage on any device other than an Ipad pro 10.5 inch.  (device id is iPad7,4)
NOTE: when running in the simulator the device is of the underlying machines processor rather than an apple device id (x86_64) and so the app will block on the simulator (which is great for displaying functionality)

## Pull Request checklist:

- [X] [WIP] tag removed from PR title
- [X] PR has an understandable description

- [X] branch name comply with our branching strategy
- [X] git branch contains relevant JIRA ticket number
- [X] branch rebased against the latest develop
- [ ] commits are squashed

- [X] Code has been tested manually
- [X] Tests are passing
- [X] PR link added to JIRA
- [X] Reviewers selected in Github

- [X] Screenshot(s) captured on the physical device (if applicable)
- [ ] Tested by QA
- [ ] PO's approval
img_a5da2357497f-1


johnmorleybjss:
Could you post a screenshot please @rossfauldsbjss?

src/environment/environment.dev.ts
```diff
@@ -3,6 +3,7 @@ import { EnvironmentFile } from './models/environment.model';
 export const environment: EnvironmentFile = {
   isRemote: true,
   configUrl: 'https://dev.mes.dev-dvsacloud.uk/v1/configuration/dev',
+  approvedDeviceIdentifiers: ['iPad7,4'],
```
johnmorleybjss: Isn't this config that is going to live in the configuration service? If so, shouldn't it be in the `AppConfig` type, but not the environment model? However, I see that you check the device type before getting configuration from the backend. Have we considered moving the config call to before the device type check?

src/providers/device/__mocks__/device.mock.ts
```diff
@@ -0,0 +1,31 @@
+import { IDeviceProvider } from '../device.model';
+import { Injectable } from '@angular/core';
+// import { Device } from '@ionic-native/device';
```
johnmorleybjss: Couple of left-over comments in this file.

src/environment/environment.dev.ts
Cavalier07: It should be in the LocalEnviromentModel and appear in local.env as we don't call out to the config service there. It doesn't need to be in any of the others.

rossfauldsbjss: it looks like I didn't fully understand how the config operated.  I do now :) so have modified accordingly.

src/providers/device/__mocks__/device.mock.ts
rossfauldsbjss: Comments removed.

src/pages/login/login.ts
```diff
@@ -49,22 +55,33 @@ export class LoginPage extends BasePageComponent {
...
-          }),
-      )
+    .then(() => {
+      this.authenticationProvider
+      .login()
+      .then(() => this.appConfigProvider.loadRemoteConfig())
+      .then(() => this.analytics.initialiseAnalytics())
+      .then(() => this.store$.dispatch(new StartSendingLogs()))
+      .then(() => {
+        const validDevice = this.deviceProvider.validDeviceType();
```
szabolcs-keresztes: This could be moved out to a function so that we keep this login function as lean as possible.

rossfauldsbjss: Great suggestion. I had done this yesterday, but had to revert and reapply and missed doing this. Have changed code now.

andrew-lai-bjss:
Blocking the application for working on a simulator will prevent the UI tests from running.

PR #188, opened 2019-03-11
mes-1739 special requirements split from comments.

Jason-Pullen:
## Pull Request checklist:

- [ ] commits are squashed

- [x] Tests are passing
- [x] PR link added to JIRA

- [ ] Tested by QA
- [ ] PO's approval

Screenshot 2019-03-11 at 12 01 29
Screenshot 2019-03-11 at 12 28 00

### DIFF
--- a/src/pages/candidate-details/__tests__/candidate-details.spec.ts
+++ b/src/pages/candidate-details/__tests__/candidate-details.spec.ts
@@ -37,14 +37,16 @@ describe('CandidateDetailsPage', () => {
     it('should create', () => {
       expect(component).toBeDefined();
     });
+  });
 
-    it('specialNeedsIsPopulated returns true for a populated array', () => {
+  describe('specialNeedsIsPopulated', () =>{
+    it('returns true for a populated array', () => {
       const specialNeedsString: string[] = ['one', 'two', 'three', 'four'];
       const result = component.specialNeedsIsPopulated(specialNeedsString);
       expect(result).toEqual(true);
     });
 
-    it('specialNeedsIsPopulated returns false for string', () => {
+    it('returns false for string', () => {
       const specialNeedsString: string = 'No details supplied';
       const result = component.specialNeedsIsPopulated(specialNeedsString);
       expect(result).toEqual(false);
--- a/src/pages/candidate-details/__tests__/candidate-details.spec.ts
+++ b/src/pages/candidate-details/__tests__/candidate-details.spec.ts
@@ -37,6 +37,18 @@ describe('CandidateDetailsPage', () => {
     it('should create', () => {
       expect(component).toBeDefined();
     });
+
+    it('specialNeedsIsPopulated returns true for a populated array', () => {
+      const specialNeedsString: string[] = ['one', 'two', 'three', 'four'];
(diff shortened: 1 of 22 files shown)


src/pages/candidate-details/candidate-details.ts
```diff
@@ -107,4 +106,8 @@ export class CandidateDetailsPage extends BasePageComponent implements OnInit, O
   handleDoneButtonClick(): void {
     this.navController.pop();
   }
+  public specialNeedsIsPopulated(specialNeeds: string | string[]): boolean {
```
rossfauldsbjss: If a string for special needs was passed instead of an array of strings it would return false for isArray.

src/pages/candidate-details/__tests__/candidate-details.spec.ts
```diff
@@ -37,6 +37,18 @@ describe('CandidateDetailsPage', () => {
     it('should create', () => {
       expect(component).toBeDefined();
     });
+
+    it('specialNeedsIsPopulated returns true for a populated array', () => {
```
szabolcs-keresztes: `specialNeedsIsPopulated` could be a `describe` which holds these 2 test cases together, so that we have a english readable sentence as a test name.

src/pages/candidate-details/candidate-details.model.ts
```diff
@@ -3,9 +3,9 @@ export type Details = {
   slotType: string,
   driverNumber: string,
   applicationRef: string,
+  specialNeeds: string | string[],
```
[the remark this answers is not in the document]
Jason-Pullen: if the formatting is the same for 'None' and actual special requirements content, we could do that. the format is slightly different so a needed a check mechanism to either check for 'None' directly(which is a little brittle), on type or set a property. Maybe I should have gone for a isPopulated property for clarity?

src/pages/candidate-details/candidate-details.selector.ts
```diff
@@ -109,3 +109,6 @@ export const getDetails = (slot: any): Details => {
   };
   return details;
 };
+export const processSpecialNeeds = (slot: any): string | string[] => {
```
szabolcs-keresztes: Could we handle `None` as a first element of `string[]`? Because in that scenario we wouldn't need to have more than one type for special needs.

src/pages/candidate-details/candidate-details.ts
Jason-Pullen: The only time it will be a string is when, the specialneeds is not populated, the selector will set that as a string 'None'. If specialneeds is a string from source, it is spit anyway turning into a single item array

src/pages/candidate-details/candidate-details.selector.ts
Jason-Pullen: Yeah I think we could, but we need to do a check to see if SN is populated, so we can format and set the 'no special needs' content and show the indicator. We could check on the first element but I feel that almost the same thing as what I'm doing.

src/pages/candidate-details/__tests__/candidate-details.spec.ts
Jason-Pullen: now nested under a describe.

PR #202, opened 2019-03-15
Feature mes 1401 view journal when not connected to network

rossfauldsbjss:
MES 1401 - Journal should be viewable from locally stored data when offline.

## Pull Request checklist:

- [x] Tests are passing
- [x] PR link added to JIRA

- [x] Tested by QA
- [x] PO's approval


src/providers/app-config/app-config.ts
```diff
@@ -30,11 +36,27 @@ export class AppConfigProvider {
...
+      } else {
+        this.getOfflineConfig()
+        .then(data => resolve(data))
+        .catch(error => reject(error));
+      }
+    })
+
+  private getOfflineConfig = (): Promise<any> =>
+    new Promise((resolve, reject) => {
+      this.dataStore.getItem('CONFIG')
```
ammarhaiderbjss: Looks like this is already a promise so doesn't need to be wrapped in one too..

src/providers/journal/journal.ts
```diff
@@ -5,28 +5,57 @@ import { UrlProvider } from '../url/url';
...
+    return this.getOfflineJournal();
+  }
+
+  getOfflineJournal(): Observable<ExaminerWorkSchedule> {
+    return from(this.getAndConvertOfflineJournal());
+  }
+
+  getAndConvertOfflineJournal(): Promise<ExaminerWorkSchedule> {
+    return new Promise<ExaminerWorkSchedule>((resolve, reject) => {
+      this.dataStore.getItem('JOURNAL').then((data) => {
```
ammarhaiderbjss: Again here, wrapping a promise in a promise

src/providers/app-config/app-config.ts
johnmorleybjss: Could we rename `getOfflineConfig` to something like `getCachedRemoteConfig` - don't want us to get confused with config that's baked into the IPA.

src/providers/data-store/data-store.ts
```diff
@@ -0,0 +1,57 @@
...
+    this.secureContainer = container;
+  }
+
+  getSecureContainer(): SecureStorageObject {
+    return this.secureContainer;
+  }
+
+  getKeys(): any {
+    if (!this.secureContainer) {
+      return Promise.resolve('getKeys');
```
johnmorleybjss: Blank array would probably be better here and then you can put a type other than `any` on the method (keeps the type uniform).

src/providers/network-state/network-state.ts
```diff
@@ -46,6 +43,9 @@ export class NetworkStateProvider {
   }
 
   public getNetworkState(): ConnectionStatus {
+    if (!this.networkStatus$) {
```
johnmorleybjss: Is this a workaround for running in browser? I don't understand the reasoning, so might be worth a comment explaining why.

src/providers/app-config/app-config.ts
rossfauldsbjss: changed

src/providers/journal/journal.ts
rossfauldsbjss: changed

src/pages/login/login.ts
```diff
@@ -44,6 +42,7 @@ export class LoginPage extends BasePageComponent {
 
     // Trigger Authentication if this isn't a logout and is an ios device
     if (!this.hasUserLoggedOut && this.isIos()) {
+      console.log('trying to login');
```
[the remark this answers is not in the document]
rossfauldsbjss: removed

src/providers/app-config/app-config.ts
rossfauldsbjss: done

src/providers/data-store/data-store.ts
rossfauldsbjss: done

src/providers/network-state/network-state.ts
rossfauldsbjss: It is - and I'm going to add some JSdocs in this file to explain.

rossfauldsbjss: Added jsdoc for this file.

Cavalier07:
Was looking at the ticket and the work looks but , however I wanted to question the approach to local data storage.

In this implementation we store each piece of data (journal and config) and mange that individually. Would a better approach be to do this more at an app level and store the redux store in the database. and when we update the store an effect is triggered to push that to local storage. When we load the app if offline we only need to pull one thing from storage not 3 or 4 different things.

With some of the upcoming tickets like recovering from a crash , reducing loss of state/data , I can see this alternative approach making our lives easier and potentially causing less work (as test-report offline will 90% be done as well) 

I wanted to open up the conversation and see what your thoughts were

ammarhaiderbjss:
> Was looking at the ticket and the work looks good , however I wanted to question the approach to local data storage.
> 
> In this implementation we store each piece of data (journal and config) and mange that individually. Would a better approach be to do this more at an app level and store the redux store in the database. and when we update the store an effect is triggered to push that to local storage. When we load the app if offline we only need to pull one thing from storage not 3 or 4 different things.
> 
> With some of the upcoming tickets like recovering from a crash , reducing loss of state/data , I can see this alternative approach making our lives easier and potentially causing less work (as test-report offline will 90% be done as well)
> 
> Work would be needed to put the config in the store but thats really where it should be anyway
> 
> I wanted to open up the conversation and see what your thoughts were

Agreed. This makes a lot of sense and something we should definitely do. However, I think it sits outside the scope of this work. The next time we save something to the device, we'll look at introducing this new generic pattern and then we can refactor the config and journal stuff to fall in line.

szabolcs-keresztes:
We already have a bunch of things to refactor. My intuition says that we should start doing things the right way from the beginning as much as we can. Because the 6 sprints wall in the office will not allow us to do refactoring any time soon.

szabolcs-keresztes:
Also would like to add that the unit testing framework what we have in place makes refactoring very challenging.

PR #207, opened 2019-03-18
MES-1801 Fixed font update bug on app first launch

roberthall12:
Fixed the bug where the font wasn't updating to the accessibility settings on first launch.
https://jira.i-env.net/browse/MES-1801

## Pull Request checklist:

- [ ] branch rebased against the latest develop
- [x] commits are squashed

- [x] Tests are passing
- [x] PR link added to JIRA

- [x] Tested by QA
- [ ] PO's approval

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -50,6 +50,7 @@ export class App {
 
   configureAccessibility() {
     if (this.platform.is('ios') && window && window.MobileAccessibility) {
+      window.MobileAccessibility.updateTextZoom();
       window.MobileAccessibility.getTextZoom(this.getTextZoomCallback);
       window.MobileAccessibility.isDarkerSystemColorsEnabled(
         (increasedContrast: boolean) => this.increasedContrast = increasedContrast);


buzzub:
I think this will conflict with the class based method of increasing text size. From what I understood, this is like pressing CMD + in Chrome and simply zooms the webview. Combined with the class based approach to increasing text size this will double the zoom effect.

This would mean that the test report page would have all elements zoomed when this is not intended.

Could you check whether this is the case?

roberthall12:
> I think this will conflict with the class based method of increasing text size. From what I understood, this is like pressing CMD + in Chrome and simply zooms the webview. Combined with the class based approach to increasing text size this will double the zoom effect.
> 
> This would mean that the test report page would have all elements zoomed when this is not intended.
> 
> Could you check whether this is the case?



> I think this will conflict with the class based method of increasing text size. From what I understood, this is like pressing CMD + in Chrome and simply zooms the webview. Combined with the class based approach to increasing text size this will double the zoom effect.
> 
> This would mean that the test report page would have all elements zoomed when this is not intended.
> 
> Could you check whether this is the case?

I don't think so, I believe it just updates the locally stored IOS zoom setting that the app uses so this shouldn't conflict with the resizing class. It was just the textZoom variable that wasn't being updated. Is there any way I can test on screens other than the journal?

buzzub:
The zoom level on the waiting room or test report page shouldn't change, so you could test that

roberthall12:
Tested on waiting room page and no change in font size. Can't seem to navigate to test report page in the app.

PR #214, opened 2019-03-19
Fix issue where preTest declarations are persisting between candidates

johnmorleybjss:
## Description and relevant Jira numbers
* Waiting room page state was propagating to different test slots when pressing back
* Ensure we clear down the `preTestDeclarations` state when "Back" is pressed
* Confirmed that "Back" will not be available beyond the Waiting Room page. As such, state accumulated later in the test can be cleared down at one of the defined test exit points (pass/fail/terminate)

## Pull Request checklist:

- [ ] [WIP] tag removed from PR title
- [ ] PR has an understandable description

## Git feature branch checklist:

- [ ] branch name comply with our branching strategy
- [ ] git branch contains relevant JIRA ticket number
- [ ] branch rebased against the latest develop
- [ ] commits are squashed

## Sign off process checklist:

- [ ] Code has been tested manually
- [ ] Tests are passing
- [ ] PR link added to JIRA
- [ ] Reviewers selected in Github
- [ ] Any new reusable component/widget added to component library on Confluence

- [ ] Screenshot(s) captured on the physical device (if applicable)
- [ ] Tested by QA
- [ ] PO's approval

PR #220, opened 2019-03-21
MES-995: Refactor state tree to nest captured data inside a particular test (MES-1921)

johnmorleybjss:
* Previously, data captured during the test was at the top level of the state (e.g. `/preTestDeclarations`)
* Introduce a top level key, `/tests`
* Under `/tests`, have `currentTest` and `startedTests` keys
* Store each individual test under `startedTests`, key them by journal `slotId` 
* Introduce a new `TestModule`, remove `CandidateModule` and `PreTestDeclarationsModule`
* Create a new `testReducer`, which wraps all test state manipulation actions and is responsible for navigating the correct test's state, then calling reducers for further manipulation

* Depends on https://github.com/dvsa/mes-api-definitions/pull/6/files for schema changes ✅
* This is the alternative approach mentioned on https://github.com/dvsa/mes-mobile-app/pull/214

image

## Pull Request checklist:

- [x] commits are squashed

- [x] Tests are passing
- [x] PR link added to JIRA

- [ ] Tested by QA
- [ ] PO's approval


ammarhaiderbjss:
@johnmorleybjss looks good so far 👍

src/modules/test/test.reducer.ts
```diff
@@ -0,0 +1,61 @@
...
+      ),
+    },
+    current: {
+      slotId,
+    },
+  };
+};
+
+const deriveSlotId = (state, action): string | null => {
+  if (action instanceof testOutcomeActions.TestOutcomeStartTest) {
```
ammarhaiderbjss: How come we've departed from the standard `switch` statement pattern for reacting to action types??

johnmorleybjss: I think the reason for this departure is that the test reducer is serving a different purpose. The rest of our reducers are "leaf" reducers and are taking actions and directly manipulating one piece of state. This is a parent reducer which delegates most of its manipulation to child reducers. I didn't see much need to `switch`, as all actions are passed to every listed reducer and _the children_ decide whether they care about that action by switching themselves. Perhaps there's a way we could refactor to make it more consistent with our other reducers?

ammarhaiderbjss: Makes perfect sense, this is a common concept in redux. Have a look at this page https://redux.js.org/recipes/structuring-reducers/splitting-reducer-logic, there's a built-in function `combineReducers` that does what you're looking for I think, i.e. splitting up reducers and having them operate on slices of nested state, looks to be available in `ngrx` also - https://ngrx.io/api/store/combineReducers

johnmorleybjss: Managed to simplify the code by using `combineReducers` for the individual test scoped state, but still need the custom wrapper around that to navigate to the correct test.

src/modules/tests/tests.reducer.ts
```diff
@@ -0,0 +1,49 @@
...
+) => {
+  const slotId = deriveSlotId(state, action);
+  if (!slotId) {
+    return state;
+  }
+
+  return {
+    ...state,
+    [slotId]: {
+      ...combineReducers(
```
ammarhaiderbjss: YES! 😍

src/modules/tests/tests.reducer.ts
```diff
@@ -0,0 +1,49 @@
...
+import * as testOutcomeActions from '../../pages/journal/components/test-outcome/test-outcome.actions';
+import { preTestDeclarationsReducer } from './pre-test-declarations/pre-test-declarations.reducer';
+import { candidateReducer } from './candidate/candidate.reducer';
+import { get } from 'lodash';
+import { combineReducers } from '@ngrx/store';
+
+// Extend this with any new test domain action types
+type TestAction = preTestDeclarationActions.Types | testOutcomeActions.Types;
+
+const initialState = {};
```
ammarhaiderbjss: Can we add some type safety to this?

johnmorleybjss: Ah yes, I knew there was another thing I wanted to do! 😁

src/modules/tests/tests.reducer.ts
ammarhaiderbjss: Can we add a special comment here to explain this esoteric usage of `combineReducers`?

szabolcs-keresztes:
Great solution :)

src/modules/tests/tests.reducer.ts
johnmorleybjss: Comment added

src/modules/tests/tests.reducer.ts
johnmorleybjss: Type safety added.

PR #274, opened 2019-04-08
Refactoring

szabolcs-keresztes:
- Renaming `fault-counter` to `driving-faults-badge`
- Replacing `TEST_OUTCOME_START_TEST` action with smaller, more accurate actions
- Removing `store$.dispatch` from `journal.effects` and just returning an array of actions instead


src/modules/tests/application-reference/application-reference.actions.ts
```diff
@@ -0,0 +1,11 @@
+import { Action } from '@ngrx/store';
+
+export const POPULATE_APPLICATION_REFERENCE = '[JournalEffects] Populate Application Reference';
+
+export class PopulateApplicationReference implements Action {
+  readonly type = POPULATE_APPLICATION_REFERENCE;
+  constructor(public payload: any) {}
```
Cavalier07: Could this be typed to Booking instead of any

johnmorleybjss: Could you use the `Application` type here?

src/modules/tests/candidate/candidate.actions.ts
```diff
@@ -0,0 +1,11 @@
+import { Action } from '@ngrx/store';
+
+export const POPULATE_CANDIDATE_DETAILS = '[JournalEffects] Populate Candidate Details';
+
+export class PopulateCandidateDetails implements Action {
+  readonly type = POPULATE_CANDIDATE_DETAILS;
+  constructor(public payload: any) {}
```
Cavalier07: Can this be type Candidate instead of any

src/pages/journal/components/test-outcome/test-outcome.ts
```diff
@@ -2,15 +2,16 @@ import { Component, Input } from '@angular/core';
...
 
 @Component({
   selector: 'test-outcome',
   templateUrl: 'test-outcome.html',
 })
 export class TestOutcomeComponent {
+
   @Input()
-  slot: any;
+  slotId: string;
```
Cavalier07: slotId is a number is the schema

src/pages/journal/__tests__/journal.effects.spec.ts
```diff
@@ -135,13 +135,16 @@ describe('Journal Effects', () => {
...
       false,
       new Date())); // Load in mock journal state
-    spyOn(store$, 'dispatch');
     // ACT
     actions$.next(new journalActions.SelectNextDay());
     // ASSERT
     effects.selectNextDayEffect$.subscribe((result) => {
-      expect(store$.dispatch).toHaveBeenCalledWith(new journalActions.JournalNavigateDay(nextDay));
-      expect(result).toEqual(new journalActions.SetSelectedDate(nextDay));
+      if (result instanceof journalActions.SetSelectedDate)  {
```
johnmorleybjss: Conditions around assertions seem a bit fishy, why are these required?

src/pages/journal/components/test-outcome/test-outcome.ts
szabolcs-keresztes: Fair point.
I'm not a big fan of having `slotId` as a number, but will change it for consistency.

src/modules/tests/application-reference/application-reference.actions.ts
szabolcs-keresztes: Will make it Application

src/modules/tests/candidate/candidate.actions.ts
szabolcs-keresztes: cool

src/pages/journal/__tests__/journal.effects.spec.ts
ammarhaiderbjss: do we need these ifs?

szabolcs-keresztes: I feel that this is a hackie solution. But couldn't do it better. So the function in subscribe will be called twice because of the elements of the action array will be arrive one by one. First time the first if will be true so we can make the right assertion and second time the second.

Let's have a think about this in the future and come up with a better solution :)

src/modules/tests/application-reference/application-reference.actions.ts
szabolcs-keresztes: just pushed the update

src/pages/journal/__tests__/journal.effects.spec.ts
szabolcs-keresztes: Agree. See my answer below

PR #279, opened 2019-04-09
MES-2047: Setup test lifecycle, moving to started once progressing beyond Waiting Room (MES-2028)

johnmorleybjss:
* Add a new key in the store under `tests` to store `testLifecycles`
* Keep this state tracking distinct from captured test data model (`startedTests`) to ensure it's not submitted to the backend
* Ensure that tests from from `NotStarted` to `Started` when Waiting Room form is submitted

**State tree**
```
tests: {
    currentTest: {
      slotId: '1005'
    },
    startedTests: {
      '1005': {
          ...
      }
    },
    testLifecycles: {
      '1005': 'Started'
    }
}
```

## Pull Request checklist:

- [ ] commits are squashed

- [x] Tests are passing
- [ ] PR link added to JIRA

- [ ] Tested by QA
- [ ] PO's approval

### DIFF
--- a/src/modules/tests/test-status/test-status.model.ts
+++ b/src/modules/tests/test-status/test-status.model.ts
@@ -1,5 +1,5 @@
 export enum TestStatus {
-  NotStarted = 'NotStarted',
+  Booked = 'Booked',
   Started = 'Started',
   Decided = 'Decided',
 }
--- a/src/modules/tests/test-status/test-status.model.ts
+++ b/src/modules/tests/test-status/test-status.model.ts
@@ -1,0 +1,5 @@
+export enum TestStatus {
+  NotStarted = 'NotStarted',
+  Started = 'Started',
+  Decided = 'Decided',
+}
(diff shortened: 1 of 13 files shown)


src/modules/tests/test-status/test-status.model.ts
```diff
@@ -0,0 +1,5 @@
+export enum TestStatus {
+  NotStarted = 'NotStarted',
```
rstepnowski: I was wondering if the status 'booked' would be more appropriate (instead of 'NotStarted') to keep it more in line with the model on wiki - if it's on DE's journal it means that it's scheduled/booked

src/modules/tests/tests.reducer.ts
```diff
@@ -20,11 +22,13 @@ export interface CurrentTest {
 export interface TestsModel {
   currentTest: CurrentTest;
   startedTests: { [slotId: string]: StandardCarTestCATBSchema };
+  testLifecycles: { [slotId: string]: TestStatus };
```
rstepnowski: Are there benefits of keeping track of test's life cycle outside of the startedTest[<SlotId>] collection?   
Maybe it would be better to introduce the test's state into the schema (for all categories)?

src/modules/tests/test-status/test-status.model.ts
johnmorleybjss: Good shout, I will use 'Booked' instead.

src/modules/tests/tests.reducer.ts
johnmorleybjss: The reason I excluded from `startedTests` is that this seemed to be tracking information specifically for the mobile app - I didn't see a reason that this should be submitted to the backend (which would imply it's part of the test schema). Do you think we would submit this status?

johnmorleybjss: Going to merge as-is for now, if we want to extend the test schema and refactor this state into there, I can do it later in this story.

PR #311, opened 2019-04-17
Recording a serious/dangerous fault on manoeuvres

szabolcs-keresztes:
- Recording a serious or a dangerous fault on manoeuvres.
- Toggling the mode of the test-report after recording a fault.
- There is possibility to record only one type of fault and if that type is a simple driving fault then it can be only one.

### DIFF
--- a/src/modules/tests/test_data/test-data.reducer.ts
+++ b/src/modules/tests/test_data/test-data.reducer.ts
@@ -33,6 +33,22 @@ export function testDataReducer(
           [action.payload]: 'DF',
         },
       };
+    case testDataActions.ADD_MANOEUVRE_SERIOUS_FAULT:
+      return {
+        ...state,
+        manoeuvres: {
+          ...state.manoeuvres,
+          [action.payload]: 'S',
+        },
+      };
+    case testDataActions.ADD_MANOEUVRE_DANGEROUS_FAULT:
+      return {
+        ...state,
+        manoeuvres: {
+          ...state.manoeuvres,
+          [action.payload]: 'D',
+        },
+      };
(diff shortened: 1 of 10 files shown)


src/modules/tests/test_data/test-data.reducer.ts
```diff
@@ -33,6 +33,22 @@ export function testDataReducer(
           [action.payload]: 'DF',
         },
       };
+    case testDataActions.ADD_MANOEUVRE_SERIOUS_FAULT:
+      return {
+        ...state,
+        manoeuvres: {
+          ...state.manoeuvres,
+          [action.payload]: 'S',
```
ammarhaiderbjss: I started us off down this road but could we utilise the `CompetencyOutcome` enum here instead of these strings?

szabolcs-keresztes: Sure, good idea!

src/modules/tests/test_data/test-data.reducer.ts
```diff
@@ -33,6 +33,22 @@ export function testDataReducer(
...
+        },
+      };
+    case testDataActions.ADD_MANOEUVRE_DANGEROUS_FAULT:
+      return {
+        ...state,
+        manoeuvres: {
+          ...state.manoeuvres,
+          [action.payload]: 'D',
```
johnmorleybjss: Would be nice to have a more descriptive name for the parameter on these actions.

szabolcs-keresztes: True, but the type of the `payload` suggests what is it going to be. Also, all `test-data.actions` have `payload` as payload name

PR #337, opened 2019-04-26
MES-2187 Adapting the the new schema

szabolcs-keresztes:
- [x] Adopting the new Vehicle Checks schema
- [x] Moving VehicleChecks state to TestData


src/pages/office/office.ts
```diff
@@ -196,10 +199,12 @@ export class OfficePage extends BasePageComponent {
         select(getAdditionalInformation),
       ),
       showMeQuestion$: currentTest$.pipe(
+        select(getTestData),
```
johnmorleybjss: As we're piping `currentTest$`, why do we need to do `getTestData` again?

src/pages/waiting-room-to-car/waiting-room-to-car.ts
```diff
@@ -178,14 +179,17 @@ export class WaitingRoomToCarPage extends BasePageComponent {
         map(isManual),
       ),
       tellMeQuestionSelected$: currentTest$.pipe(
+        select(getTestData),
```
johnmorleybjss: Same as on Office, we're doing `currentTest$.pipe`, wouldn't have thought that `getTestData` was necessary.

src/pages/office/office.ts
szabolcs-keresztes: Good point. Weird that it works fine..
Will change it.

szabolcs-keresztes: No, you need the `getTestData` selector. The vehicle checks are part of the `testData`

johnmorleybjss: Ah yes, I understand now, `testData` is the key within the individual test, fair enough. 😄

PR #338, opened 2019-04-26
MES-2162: Repopulate serious and dangerous fault comments when returning to Office (MES-1129)

johnmorleybjss:
* Create `FaultCommentCardComponent` and `FaultCommentComponent` component
* Allow each `FaultCommentComponent` to manage its own `FormControl` in the parent `FormGroup` and patch the value when there's a change in the store
* Remove lots of now redundant code in the Office page
* Remove fault type specific comment components

## Pull Request checklist:

- [ ] commits are squashed

- [x] Tests are passing
- [x] PR link added to JIRA

- [ ] Tested by QA
- [ ] PO's approval


src/pages/debrief/debrief.selector.ts
```diff
@@ -73,16 +75,14 @@ export const displayDrivingFaultComments = (data: TestData) : boolean => {
  *
  * Returns a container of array holding the propertyName and fullCompetencyLabel for each dangerous
  * fault recorded against a candidate.
- *
- * Is used to populate properties in dangerous-fault-comment.html for the entry of comments
- * relating to each dangerous fault.
  */
 export const getDangerousFaults = (faults: DangerousFaults): SeriousFaultsContainer[] => {
   const faultsEncountered: SeriousFaultsContainer[] = [];
```
rossfauldsbjss: Don't like the naming of SeriousFaultscontainer - when used for getting dangerous faults. 
Probably my mistake in providing something call getSeriousOrDangerousFaults and not wanting to duplicate code, but would prefer a more generic name for this as it looks wrong that getDangerousFaults returns a SeriousFaultsContainer

johnmorleybjss: Yes I agree, `SeriousFaultsContainer` is from the original implementation, I'm hoping to rationalise that and `FaultComment` before merging. 👍

buzzub: +1 for ngSwitch. Seems like a good use case for it.

src/pages/office/components/fault-comment/fault-comment.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    this.formControl.patchValue(this.faultComment.comment);
+  }
+
+  faultCommentChanged(comment: string): void {
+    const faultComment: FaultComment = { comment, competency: this.faultComment.propertyName };
+    this.faultCommentChange.emit(faultComment);
+  }
+
+  get invalid() {
```
szabolcs-keresztes: `isValid` getter would sound better and maybe negate where ever it's need to be checked if is invalid

johnmorleybjss: I agree, however, this would mean changing it in the rest of the components too. Seeing as this is working and fulfilling the user need as it is, I will look into doing this in another PR with some other tweaks.

PR #353, opened 2019-05-01
MES-1013 Vehicle Checks

szabolcs-keresztes:
- Record a dangerous fault against show me question
- Record a driving fault agains show me question
- Toggle show me question's completed status
- Remove show me question faults
- Update global fault counter

### DIFF
--- a/src/modules/tests/test-data/test-data.reducer.ts
+++ b/src/modules/tests/test-data/test-data.reducer.ts
@@ -323,6 +323,17 @@ export function testDataReducer(
           },
         },
       };
+    case testDataActions.SHOW_ME_QUESTION_DRIVING_FAULT:
+      return {
+        ...state,
+        vehicleChecks: {
+          ...state.vehicleChecks,
+          showMeQuestion: {
+            ...state.vehicleChecks.showMeQuestion,
+            outcome: CompetencyOutcome.DF,
+          },
+        },
+      };
     default:
       return state;
   }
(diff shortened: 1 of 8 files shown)


src/modules/tests/test-data/test-data.reducer.ts
```diff
@@ -312,6 +323,39 @@ export function testDataReducer(
...
+        },
+      };
+    case testDataActions.SHOW_ME_QUESTION_REMOVE_FAULT:
+      return {
+        ...state,
+        vehicleChecks: {
+          ...state.vehicleChecks,
+          showMeQuestion: {
+            ...state.vehicleChecks.showMeQuestion,
+            outcome: undefined,
```
ammarhaiderbjss: Might need to check with @ali-campbell but I think we should be removing the outcome when it is removed..

ammarhaiderbjss:
@szabolcs-keresztes Awesome, very comprehensive and thorough tests 😍, just a query on the removing fault reducer then happy to approve.

src/modules/tests/test-data/test-data.reducer.ts
szabolcs-keresztes: Okay, I am going to remove it

ali-campbell: Yes, if there's no need to know that a fault was removed (e.g. for audit) then we should remove the outcome.  Saves on storage and easier for validation.

PR #364, opened 2019-05-03
Spike mes 2067 show fields based on activity code

rossfauldsbjss:
MES 2067 is a tech spike to prove that we can map the data from the validation spreadsheet to a data structure and use that data to control displaying form fields (or not) and validation of those fields (or not), plus changing options in drop downs etc.

This behaviour mapping will be extended next sprint for all form controls on the office page, and for all test outcomes. 

It also relies on testOutcome (activity code) being populated so at the moment defaults to a known value.


src/providers/outcome-behaviour-map/outcome-behaviour-map.model.ts
```diff
@@ -0,0 +1,12 @@
+export interface FieldDetail {
+  display: string;
```
ammarhaiderbjss: can we make this a string union type if there are just 3 values? i.e.
`display: 'Y' | 'N' | 'A';`

src/modules/tests/__tests__/tests.selector.spec.ts
```diff
@@ -156,8 +156,8 @@ describe('testsSelector', () => {
     });
     it('should return mes-red class for non listed activity code 86', () => {
       testState.activityCode = '86';
-      const result = getTestOutcome(testState);
-      expect(result).toBe('Terminated');
+      const result = getTestOutcomeClass(testState);
+      expect(result).toBe('mes-red');
     });
   });
```
ammarhaiderbjss: Unit test for `getShowMeQuestionOptions`?

src/pages/office/office.ts
```diff
@@ -198,6 +222,9 @@ export class OfficePage extends BasePageComponent {
         select(getVehicleChecks),
         select(getShowMeQuestion),
       ),
+      showMeQuestionOptions$: currentTest$.pipe(
+        map(() => getShowMeQuestionOptions(this.showMeQuestions, '4', this.outcomeBehaviourProvider)),
```
ammarhaiderbjss: Perhaps a TODO comment here to replace the hardcoded outcome?

src/providers/outcome-behaviour-map/outcome-behaviour-map.model.ts
rossfauldsbjss: Done.

src/modules/tests/__tests__/tests.selector.spec.ts
rossfauldsbjss: oh yes. good spot. will do that now.

src/pages/office/office.ts
rossfauldsbjss: definite TODO needed. Changing now.

rossfauldsbjss: Have wired this up to take outcome from state, so no TODO needed

src/providers/outcome-behaviour-map/outcome-behaviour-map.model.ts
```diff
@@ -0,0 +1,12 @@
+export interface FieldDetail {
+  display: 'A' | 'Y' | 'N';
```
johnmorleybjss: As discussed, enum around this would be good.

PR #382, opened 2019-05-09
Mes 2271 access practice mode

buzzub:
- Add a button to the journal screen to access the practice test report (will move onto home screen in future sprint)
- Create a modal for adding driving fault
- Create method of creating a practice test / slot
- Added StartPracticeTest action to tests actions (future PR to move StartTest and ActivateTest here too)

Screenshot 2019-05-09 at 11 53 55

Screenshot 2019-05-09 at 14 41 40


buzzub:
Agreed with Tom to move `StartPracticeTest` action into `modules/tests/tests.actions.ts` and create another PR to move `StartTest` and `ActivateTest` actions here too.

Will move effect there too.

src/modules/tests/tests.effects.ts
```diff
@@ -36,4 +40,79 @@ export class TestsEffects {
...
+    ofType(testActions.START_PRACTICE_TEST),
+    switchMap(() => {
+      const slot = {
+        slotData: {
+          ...practiceSlot,
+        },
+      };
+
+      return [
+        new PopulateApplicationReference(slot.slotData.booking.application),
```
johnmorleybjss: Could you avoid creating `slot` with `slotData` property and just navigate `practiceSlot` directly?

buzzub: Yes, nice spot. This was leftover from when this lived in the journal effects class and before I realised I didn't need to initialise quite as much state. `extractTestSlotAttributes` and `extractTestCentre` are also redundant now.

src/modules/tests/tests.effects.ts
```diff
@@ -36,4 +40,60 @@ export class TestsEffects {
...
+      return [
+        new PopulateApplicationReference(slotData.booking.application),
+        new PopulateCandidateDetails(slotData.booking.candidate),
+      ];
+    }),
+  );
+
 }
+
+export const application: Application = {
```
ammarhaiderbjss: Would these be better placed in a `/mocks` folder?

buzzub: Yeah, I totally couldn't decide the best place for them. We don't have a mocks folder yet do we?

buzzub: Ah, `/__mocks__` you mean?

ammarhaiderbjss: sorry yes that's it

PR #399, opened 2019-05-14
Addition of ui tests

andrew-lai-bjss:
Adding a basic full end to end test journey as a basis for all other UI tests.

Also includes basic validation checks for Waiting room to car page


src/pages/back-to-office/back-to-office.html
```diff
@@ -1,9 +1,9 @@
-<ion-content no-bounce>
+<ion-content no-bounce id="continue-to-write-up-page">
```
johnmorleybjss: Might be nice for the ID to be `back-to-office-page` to stay consistent with the filename.

andrew-lai-bjss: Updated as suggested.

PR #436, opened 2019-05-21
MES-1975: Introducing test report & end 2 end practice mode mock data

szabolcs-keresztes:
- Adding test report practice mode mock data
- Adding end 2 end practice mode mock data
- Using the test report practice mode mock data at the relevant places

### DIFF
--- a/src/modules/tests/__mocks__/tests.mock.ts
+++ b/src/modules/tests/__mocks__/tests.mock.ts
@@ -1,10 +1,9 @@
-import { Candidate } from '@dvsa/mes-test-schema/categories/B';
-import { Application } from '../../../shared/models/DJournal';
+import { Application, Candidate } from '../../../shared/models/DJournal';
 
-export const testReportPracticeSlotId = 'test_report_practice';
-export const end2endPracticeSlotId = 'end_2_end_practice';
+export const testReportPracticeSlotId = 'practice_test_report';
+export const end2endPracticeSlotId = 'practice_end_2_end';
 
-export const testReportPracticeModeApplication: Application = {
+export const testApplicationMock: Application = {
   applicationId: 12345692,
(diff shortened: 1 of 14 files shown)


src/modules/tests/__mocks__/tests.mock.ts
```diff
@@ -24,20 +68,20 @@ export const candidate: Candidate = {
   candidateName: {
     firstName: 'Practice',
     lastName: 'Mode',
-    title: 'Miss',
+    title: 'End to End',
```
Cavalier07: We don't usually show candidate name with the title so if this is something you want to display it needs to be first or last name

src/pages/test-report/test-report.effects.ts
```diff
@@ -107,10 +107,10 @@ export class TestReportEffects {
     withLatestFrom(
       this.store$.pipe(
         select(getTests),
-        map(isPracticeTest),
+        map(isTestReportPracticeTest),
       ),
     ),
-    filter(([action, isPracticeTest]) => !isPracticeTest),
+    filter(([action, isTestReportPracticeTest]) => !isTestReportPracticeTest),
```
Cavalier07: Does this mean we are going to need a double filter everywhere we want to not do something if it's practice mode (or double up on if statements) , seems like we would want to keep an isPracticeMode selector which can tell us this instead of separating them out

Cavalier07:
Can we chat about this in the morning :)

src/pages/test-report/test-report.effects.ts
szabolcs-keresztes: We have to identify the different practice modes so that we know when to exit each of them. However on the presentational layer we will have only one variable which indicates if it's practice mode not, but we don't need to know which.

PR #444, opened 2019-05-23
Mes-1980 - Adds ability to check for practice mode on all tests pages

Cavalier07:

- Adds new selectors for checking if we are in practice mode
- Adds a new base page class for pages using practice mode
- Updates all test pages to use new base class

## Pull Request checklist:

- [ ] commits are squashed

- [x] Tests are passing
- [x] PR link added to JIRA

- [ ] Tested by QA
- [ ] PO's approval

### DIFF
--- a/src/shared/classes/practiceable-base-page.ts
+++ b/src/shared/classes/practiceable-base-page.ts
@@ -18,17 +18,17 @@ interface PracticeableBasePageState {
 
 export abstract class PracticeableBasePageComponent extends BasePageComponent implements OnInit, OnDestroy {
 
-  practiceableBasePageState: PracticeableBasePageState;
-  subscription: Subscription;
+  public isPracticeMode: boolean;
 
-  isPracticeMode: boolean;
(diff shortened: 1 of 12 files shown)


src/shared/classes/practiceable-base-page.ts
```diff
@@ -0,0 +1,60 @@
+import { Platform, NavController } from 'ionic-angular';
```
ammarhaiderbjss: Is the purpose of this new base page simply to house the practice mode banner?

Cavalier07: It's an abstract class so it can't house the banner html. It was so we only have to have the logic/selecting the data for checking if it's practice mode in once place instead of in every page of the app.

The reason for not using the Base Page is we have a number of pages which won't show the banner (Login, Journal, Candidate Details)

Cavalier07: By the end of this ticket there will be other things we will want to do across all the pages , which we should be able to do from here.

PR #449, opened 2019-05-24
MES-1882: Handle HTTP 304 when fetching journals

johnmorleybjss:
* Journal service returns HTTP 304 when there are no changes as per https://github.com/dvsa/mes-journal-service/pull/28.
* Handle this gracefully by using the journal slots already cached on the device.
* Simplify how the `If-Modified-Since` header on the journal request is generated.

## Pull Request checklist:

- [ ] commits are squashed

- [x] Tests are passing
- [ ] PR link added to JIRA

- [ ] Tested by QA
- [ ] PO's approval

### DIFF
--- a/src/pages/journal/journal.effects.ts
+++ b/src/pages/journal/journal.effects.ts
@@ -88,6 +88,7 @@ export class JournalEffects {
                   lastRefreshed,
                 ));
               }
+              return of(err);
             }),
           );
       }),
--- a/src/pages/journal/journal.effects.ts
+++ b/src/pages/journal/journal.effects.ts
@@ -68,13 +68,27 @@ export class JournalEffects {
             tap((journalData: ExaminerWorkSchedule) => this.journalProvider.saveJournalForOffline(journalData)),
             map((journalData: ExaminerWorkSchedule) => this.slotProvider.detectSlotChanges(slots, journalData)),
             map((slots: any[]) => groupBy(slots, this.slotProvider.getSlotDate)),
-            map((slots: {[k: string]: SlotItem[]}) => this.slotProvider.extendWithEmptyDays(slots)),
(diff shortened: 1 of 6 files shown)


src/providers/journal/journal.ts
```diff
@@ -26,21 +26,21 @@ export class JournalProvider {
...
     private dateTimeProvider: DateTimeProvider,
-  ) {}
+  ) { }
 
   getJournal(lastRefreshed: Date): Observable<ExaminerWorkSchedule> {
     const staffNumber = this.authProvider.getEmployeeId();
     const journalUrl = this.urlProvider.getPersonalJournalUrl(staffNumber);
     const networkStatus = this.networkStateProvider.getNetworkState();
     if (lastRefreshed === null) {
       if (!this.authProvider.isInUnAuthenticatedMode() &&
```
MarkPatt89: At risk of being pernickety here, isInUnAuthenticatedMode, drop the upper case 'A'.

src/pages/journal/journal.effects.ts
```diff
@@ -68,13 +68,28 @@ export class JournalEffects {
...
+              new journalActions.LoadJournalSuccess(
+                slots,
+                this.networkStateProvider.getNetworkState(),
+                this.authProvider.isInUnAuthenticatedMode(),
+                lastRefreshed,
+              ),
+            ),
+            catchError((err) => {
+              // For HTTP 304 NOT_MODIFIED we just use the slots we already have cached
+              if (err.status === 304) {
```
Cavalier07: If we aren't modifying the data do we need to run the success action as the data will just be the same right

johnmorleybjss: I tried this, but the spinner stayed up as that action is triggering the spinner dismissal.

Cavalier07: Just looked and it seems a pain to untangle, doing it this way won't cause any issues :)

PR #450, opened 2019-05-24
Adding test for meeting place comments, and checks for CCP default an…

andrew-lai-bjss:
Updates to UI tests to add Meeting place checks and update to CCP email checks.


andrew-lai-bjss:
I also updated the newCommandTimeout so that the appium session does not expire too quickly especially when timeout events occur.

PR #466, opened 2019-05-30
Configuring the right start date for fake journal slots

szabolcs-keresztes:
- Getting today's 8am date and time
- Setting each test slot one after another with one hour difference
- Fixing the Test Report Practice mode entry button


src/pages/fake-journal/fake-journal.ts
```diff
@@ -32,4 +33,6 @@ export class FakeJournalPage extends BasePageComponent {
     this.dateToDisplay = moment().format('dddd D MMMM YYYY');
   }
 
+  isRedLineSlot = (slotId: string) => slotId === `${end2endPracticeSlotId}_2`;
```
Cavalier07: Can we change this so it's easier to understand

PR #470, opened 2019-05-31
Feature mes 2368 unauthenticated logging

rossfauldsbjss:
MES - 2368 log unauthenticated requests - e.g user cancelled sign in, or failed authorisation.
Logging service was changed by another request to use API key authentication instead of the custom authoriser using JWT.

logging API key, URL and interval are now going to be provided by MDM/airwatch. 
As the app cant access remote config if the user is unauthenticated.
Separate piece of work will remove logging parameters from remote config.

## Pull Request checklist:

- [ ] commits are squashed

- [x] Tests are passing
- [x] PR link added to JIRA


src/environment/environment.dev.ts
```diff
@@ -7,6 +7,8 @@ export const environment: EnvironmentFile = {
   daysToCacheLogs: 7,
   logoutClearsTestPersistence: true,
   logsPostApiKey: '',
+  logsApiUrl: 'https://john.mes.dev-dvsacloud.uk/v1/logs',
```
johnmorleybjss: Using my API subdomain for quite a few envs. I won't block as I'm out this afternoon, but please fix!

rossfauldsbjss: replaced john with dev in two envs.

src/environment/environment.production.ts
```diff
@@ -6,6 +6,8 @@ export const environment: EnvironmentFile = {
   daysToCacheJournalData: 7,
   daysToCacheLogs: 7,
   logsPostApiKey: '',
+  logsApiUrl: 'https://prod.mes.dev-dvsacloud.uk/v1/logs',
```
rstepnowski: The URL is incorrect for PROD but the bigger issue here is that we probably don't want to put URLs to prod env in github at all.

Given the fact that this info comes form MDM anyways - I'd prefer to leave this empty - at least for prod/uat/perf/staging(?) - is there any value of hardcoding those URLs on higher ENVs than local/dev?
If we really need those URLs documented somewhere confluence would be a better place for it imho.

rossfauldsbjss: I can remove this (and will)   - there is a url in there for the config service already though.

src/pages/login/login.ts
```diff
@@ -195,4 +217,13 @@ export class LoginPage extends BasePageComponent {
       this.loadingSpinner = null;
     }
   }
+
+  private createLog(logType: LogType, actionType: string, message: string): Log {
```
jameshighfield: Unused param `actionType`. Did you intend on adding this to the log?

rossfauldsbjss: Great spot. Unintentional. Removing superfluous parameter now.

src/pages/login/login.ts
```diff
@@ -106,8 +109,27 @@ export class LoginPage extends BasePageComponent {
...
         this.analytics.logException(error, true);
+        const log: Log = this.createLog(LogType.INFO,
+          `user cancelled login`);
+        this.store$.dispatch(new SaveLog(log));
+        this.store$.dispatch(new SendLogs());
       }
       if (error === AuthenticationError.USER_NOT_AUTHORISED) {
+
+        const token = await this.authenticationProvider.getAuthenticationToken();
+        const examiner = await this.authenticationProvider.getEmployeeId() || '';
```
ammarhaiderbjss: Perhaps change this to say 'unavailable' so the log is more intelligible?

src/pages/login/login.ts
```diff
@@ -106,8 +109,27 @@ export class LoginPage extends BasePageComponent {
 
       if (error === AuthenticationError.USER_CANCELLED) {
         this.analytics.logException(error, true);
+        const log: Log = this.createLog(LogType.INFO,
```
ammarhaiderbjss: Looks like we're repeating this a few times. Perhaps a function to handle the creation of log and dispatching called `saveAndDispatchLog` that accepts a message as a param?

src/pages/login/login.ts
rossfauldsbjss: changed.

rossfauldsbjss: done

src/pages/login/login.ts
rossfauldsbjss: done

PR #479, opened 2019-06-03
MES-2488: Renaming termination code to activity code

szabolcs-keresztes:
It's just a refactoring PR which will rename termination code to activity code.


src/pages/office/components/activity-code/activity-code.html
```diff
@@ -0,0 +1,22 @@
+<ion-row class="mes-validated-row mes-row-separator" id="activity-code-card" [formGroup]="formGroup">
```
johnmorleybjss: Would have been nice to `git mv` this file to maintain the history on it. Using the filesystem and re-adding lists it as a completely new file.

szabolcs-keresztes: True, will remember this when something similar comes up in the future

PR #480, opened 2019-06-04
Ensure that examiner staff number is copied into test state to make test submission valid

johnmorleybjss:
* Relates to https://jira.i-env.net/browse/MES-2295
* `staffNumber` needs to be included in the submission payload to be saved into the Aurora record
* Ensure that we populate it when starting a test
* **Needs confirmation of behaviour when we start a test while in unauthenticated mode**

## Pull Request checklist:

- [x] commits are squashed

- [x] Tests are passing
- [ ] PR link added to JIRA

- [ ] Tested by QA
- [ ] PO's approval

### DIFF
--- a/src/pages/journal/journal.effects.ts
+++ b/src/pages/journal/journal.effects.ts
@@ -35,6 +35,7 @@ import {
 import { PopulateTestCentre } from '../../modules/tests/test-centre/test-centre.actions';
 import { SetTestStatusBooked } from '../../modules/tests/test-status/test-status.actions';
 import { extractTestSlotAttributes } from '../../modules/tests/test-slot-attributes/test-slot-attributes.selector';
+import { PopulateExaminer } from '../../modules/tests/examiner/examiner.actions';
 
 @Injectable()
 export class JournalEffects {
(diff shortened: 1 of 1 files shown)


ammarhaiderbjss:
> Needs confirmation of behaviour when we start a test while in unauthenticated mode
Does this mean we're not sure how this is meant to work in unauthenticated mode?

johnmorleybjss:
Yes @ammarhaiderbjss, this isn't tested yet. It could be that we need to start maintaining the staff number in the journal for subsequent copy into the test - this would require poller changes. Was thinking that we could possibly merge this to get us moving again on submission for the meantime.

ammarhaiderbjss:
> Yes @ammarhaiderbjss, this isn't tested yet. It could be that we need to start maintaining the staff number in the journal for subsequent copy into the test - this would require poller changes. Was thinking that we could possibly merge this to get us moving again on submission for the meantime.

Ok cool happy to approve, re maintaining staff number from journal: that's an interesting one. I would have considered the staff number from Azure AD rather than the journal to be the source of truth when it comes to identifying who submitted a test.

johnmorleybjss:
Discussed with @rstepnowski, we are planning to annotate the journal with the logged in staff number and store it in the app state at the time of retrieval from the backend. Unless we can retrieve the staff number from authenticator while in unauthenticated mode. This needs more investigation, so a ticket will be created to address. Merging this interim solution to get the online submission working for now.

PR #496, opened 2019-06-11
Mes 2539 pdf generate spike [Spike, please don't merge]

amobarak:
To run the spike, uncomment app.component.ts:17.


src/pages/poc/poc.template.ts
```diff
@@ -0,0 +1,9 @@
+export const template: string = `
```
amobarak: This is the HTML template that's used to generate the PDF

PR #510, opened 2019-06-21
Mes 2179 addition of full journey ui tests

andrew-lai-bjss:
Addition of extra full journey UI tests. 

Also adds in extended regression tests which are to be executed against the release candidate but does not form part of the standard build pipeline. This is the ensure that the pipeline UI test do not exceed 30 minutes but offer a good spread of test coverage offering a high level of confidence in the build.

### DIFF
--- a/test/e2e/extended/cat-b-scenarios.feature
+++ b/test/e2e/extended/cat-b-scenarios.feature
@@ -106,6 +106,54 @@ Feature: Extended category B test scenarios
       And I upload the test
       Then I should see the "Journal" page
 
+   Scenario: Candidate passes a test with 15 driver faults
+      Given I am logged in as "mobexaminer1" and I have a test for "Mrs Jane Doe"
+      When I start the test for "Mrs Jane Doe"
+      And the candidate enters a new email address
+      And the candidate confirms their communication preference
+      Then I should see the "Declaration - Jane Doe" page
+      And the candidate completes the declaration page
+      And I proceed to the car
+      Then I should see the "Jane Doe" page
+      And I complete the waiting room to car page
+      Then I should see the "Test report - Jane Doe" page
+      When I add a "Accelerator" driver fault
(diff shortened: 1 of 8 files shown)


test/e2e/extended/cat-b-scenarios.feature
```diff
@@ -0,0 +1,198 @@
...
+      When I end the debrief
+      Then I should see the "Test debrief - Florence Pearson" page
+      And I complete the pass details
+      And I complete the health declaration
+      Then I am on the back to office page
+      And I continue to the office write up
+      Then I should see the "Office" page
+      And I complete the office write up
+      And I upload the test
+      Then I should see the "Journal" page
```
johnmorleybjss: Wondering whether we can do anything here around confirming that the upload is successful, but this will probably require some additional work to clear out potential duplicates. Worth some thought.

test/e2e/extended/cat-b-scenarios.feature
```diff
@@ -0,0 +1,198 @@
...
+      Then I should see the "Declaration - Jane Doe" page
+      And the candidate completes the declaration page
+      And I proceed to the car
+      Then I should see the "Jane Doe" page
+      And I complete the waiting room to car page with a tell me driver fault
+      Then I should see the "Test report - Jane Doe" page
+      When I add a "Accelerator" serious fault
+      When I complete the test
+      And I continue to debrief
+      Then I should see the Debrief page with outcome "Unsuccessful"
```
johnmorleybjss: Maybe some assertion about the list of faults? I see we have this for the full journey suite, so would have thought they'd be included in extended suite.

test/e2e/extended/cat-b-scenarios.feature
andrew-lai-bjss: Yep agreed, adds value moving forwards and will look to implement something in a future change.

PR #521, opened 2019-06-27
Hook up provider and ui

szabolcs-keresztes:
Use Candidate details search to make a DB query.
Logging out the array.


src/pages/test-results-search/test-results-search.ts
```diff
@@ -2,6 +2,12 @@ import { Component } from '@angular/core';
 import { IonicPage, NavController, NavParams, Platform } from 'ionic-angular';
 import { BasePageComponent } from '../../shared/classes/base-page';
 import { AuthenticationProvider } from '../../providers/authentication/authentication';
+import { SearchProvider } from '../../providers/search/search';
+import { tap, catchError } from 'rxjs/operators';
+import { of } from 'rxjs/observable/of';
+
+const SEARCH_BY_DRIVER_NUMBER = 'driverNumber';
```
ammarhaiderbjss: This could be an enum called `DriverDetails` and then the `searchBy` type could be changed from string to DriverDetails.

PR #539, opened 2019-07-02
MES-2864 Fix navigation issue on prod builds

buzzub:
The minification during the build process meant that the method used to find views and remove them from the nav stack failed :(

This PR switches the method of finding the view to use the `id` property of the view instead of the `name`.

In addition, there are extra checks before removing the view to ensure that if the view isn't found, it doesn't call `removeView` with a null value (doing this will remove the first page from the nav stack - which is not good).

## Pull Request checklist:

- [x] commits are squashed

- [x] Tests are passing
- [ ] PR link added to JIRA

- [ ] Tested by QA
- [ ] PO's approval

### DIFF
--- a/src/pages/debrief/debrief.ts
+++ b/src/pages/debrief/debrief.ts
@@ -239,11 +239,11 @@ export class DebriefPage extends PracticeableBasePageComponent {
       return;
     }
     this.navController.push(BACK_TO_OFFICE_PAGE).then(() => {
-      const testReportPage = this.navController.getViews().find(view => view.name === TEST_REPORT_PAGE);
+      const testReportPage = this.navController.getViews().find(view => view.id === TEST_REPORT_PAGE);
       if (testReportPage) {
         this.navController.removeView(testReportPage);
       }
-      const debriefPage = this.navController.getViews().find(view => view.name === DEBRIEF_PAGE);
+      const debriefPage = this.navController.getViews().find(view => view.id === DEBRIEF_PAGE);
       if (debriefPage) {
         this.navController.removeView(debriefPage);
       }
--- a/src/pages/debrief/debrief.ts
+++ b/src/pages/debrief/debrief.ts
@@ -243,9 +243,10 @@ export class DebriefPage extends PracticeableBasePageComponent {
       if (testReportPage) {
         this.navController.removeView(testReportPage);
       }
-      this.navController.removeView(
-        this.navController.getViews().find(view => view.name === DEBRIEF_PAGE),
-      );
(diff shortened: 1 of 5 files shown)


src/pages/debrief/debrief.ts
```diff
@@ -239,13 +239,14 @@ export class DebriefPage extends PracticeableBasePageComponent {
       return;
     }
     this.navController.push(BACK_TO_OFFICE_PAGE).then(() => {
-      const testReportPage = this.navController.getViews().find(view => view.name === TEST_REPORT_PAGE);
+      const testReportPage = this.navController.getViews().find(view => view.id === TEST_REPORT_PAGE);
```
ammarhaiderbjss: Is there any documentation on what this `id` property actually relate to? I couldn't find anything on https://ionicframework.com/docs/v3/api/navigation/ViewController/

src/pages/waiting-room-to-car/waiting-room-to-car.ts
```diff
@@ -276,9 +276,10 @@ export class WaitingRoomToCarPage extends PracticeableBasePageComponent {
     if (this.form.valid) {
       this.navController.push(TEST_REPORT_PAGE).then(() => {
         // remove Waiting Room To Car Page
-        this.navController.removeView(
-          this.navController.getViews().find(view => view.name === WAITING_ROOM_TO_CAR_PAGE),
-        );
+        const view = this.navController.getViews().find(view => view.id === WAITING_ROOM_TO_CAR_PAGE);
+        if (view) {
```
ammarhaiderbjss: How come we have to check if it exists? Should we handle the case where it doesn't?

buzzub: In this scenario the page _should_ exist as we're navigating off it, but the conditional handles the case when we can't find the view in the nav stack e.g. because the `id` or `name` property is no longer available or minified.

src/pages/debrief/debrief.ts
buzzub: It's undocumented, but seems to be the only way to find a view in the stack without resorting to counting back through indexes. If we don't use this, I think we need to abandon the idea of popping pages off and rely on `popToRoot` clearing up after the write up.

In the event that `id` is removed or changed (unlikely?) then the additional checking means that we simply wouldn't remove any pages.

andrew-lai-bjss:
Green build - https://jenkins.mgmt.mes.dvsacloud.uk/job/mobile-app_build/604/

PR #548, opened 2019-07-05
MES-1982: Post-test Analytics

szabolcs-keresztes:
- [x] Adding analytics for screen tracking
- [x] Happy path post-test analytics
- [x] Errors post-test analytics

### DIFF
--- a/src/modules/tests/tests.analytics.effects.ts
+++ b/src/modules/tests/tests.analytics.effects.ts
@@ -14,6 +14,7 @@ import {
 import { getTestById, isPassed } from './tests.selector';
 import { SEND_COMPLETED_TESTS_FAILURE, SendCompletedTestsFailure } from './tests.actions';
 import { of } from 'rxjs/observable/of';
+import { TestsModel } from './tests.model';
 
 @Injectable()
 export class TestsAnalyticsEffects {
@@ -38,9 +39,8 @@ export class TestsAnalyticsEffects {
         select(getTests),
       ),
     ),
-    concatMap(([action, tests]) => {
-      const setTestStatusSubmittedAction = action as SetTestStatusSubmitted;
-      const test = getTestById(tests, setTestStatusSubmittedAction.slotId);
+    concatMap(([action, tests]: [SetTestStatusSubmitted, TestsModel]) => {
+      const test = getTestById(tests, action.slotId);
       const isTestPassed = isPassed(test);
       const journalDataOfTest = test.journalData;
 
--- a/src/modules/tests/tests.analytics.effects.ts
+++ b/src/modules/tests/tests.analytics.effects.ts
@@ -1,0 +1,59 @@
+import { Injectable } from '@angular/core';
+import { AnalyticsProvider } from '../../providers/analytics/analytics';
+import { Actions, Effect, ofType } from '@ngrx/effects';
+import { Store, select } from '@ngrx/store';
+import { StoreModel } from '../../shared/models/store.model';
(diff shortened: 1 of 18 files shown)


src/pages/office/office.ts
```diff
@@ -542,6 +542,7 @@ export class OfficePage extends PracticeableBasePageComponent {
 
   defer() {
     this.popToRoot();
+    // this.store$.dispatch(new DeferWriteup());
```
ammarhaiderbjss: Do we not need this anymore?

src/modules/tests/tests.analytics.effects.ts
```diff
@@ -0,0 +1,59 @@
...
+      const journalDataOfTest = test.journalData;
+
+      this.analytics.logEvent(
+        AnalyticsEventCategories.POST_TEST,
+        AnalyticsEvents.SUBMIT_TEST,
+        isTestPassed ? 'pass' : 'fail',
+      );
+
+      this.analytics.addCustomDimension(
+        AnalyticsDimensionIndices.TEST_ID, journalDataOfTest.testSlotAttributes.slotId.toString());
```
buzzub: Could you use the `getCurrentTestSlotId` selector for this?

buzzub: Ah maybe not... that selects `tests.currentTest.slotId`

buzzub: Oh, you could just select `slotId` earlier and then reuse
`const slotId = getCurrentTestSlotId(tests);`

src/modules/tests/tests.selector.ts
```diff
@@ -44,7 +48,7 @@ export const isTestOutcomeSet = (test: StandardCarTestCATBSchema) => {
   return false;
 };
 
-export const isPassed = (test: StandardCarTestCATBSchema) => {
+export const isPassed = (test: StandardCarTestCATBSchema): boolean => {
```
buzzub: Niiice

src/pages/back-to-office/back-to-office.analytics.effects.ts
```diff
@@ -32,4 +38,34 @@ export class BackToOfficeAnalyticsEffects {
...
+        select(getCurrentTest),
+        select(isPassed),
+      ),
+      this.store$.pipe(
+        select(getTests),
+        select(getCurrentTest),
+        select(getJournalData),
+      ),
+    ),
+    switchMap(([action, isPassed, journalDataOfTest]) => {
```
buzzub: Is it worth specifying the type of these arguments? e.g.
`switchMap(([action, isPassed, journalDataOfTest]: [DeferWriteUp, boolean, JournalData]) => {`

src/modules/tests/tests.analytics.effects.ts
szabolcs-keresztes: The only thing with the `getCurrentTestSlotId` function is that it always gets the current test slot id. And there is a chance here that we are not setting the status of the current test, but an older test which failed to get submitted earlier. 
And the most easiest way is to go check what's in `journalData` and set that `slotId`.

src/pages/back-to-office/back-to-office.analytics.effects.ts
szabolcs-keresztes: Can do

src/pages/office/office.ts
szabolcs-keresztes: changed it

PR #554, opened 2019-07-08
MES-992 - Show Outcome Code in Journal

roberthall12:
##  Description and relevant Jira numbers

- Used protected type `SlotDetail` instead of `any` for `test-outcome` component
- Pulled through the test data into the `test-outcome` component using the `tests-selector`
- `ngIf` on whether the test is `Completed` or `Submitted` to display the activity code

https://jira.i-env.net/browse/MES-992

## Pull Request checklist:

- [ ] commits are squashed

- [x] Tests are passing
- [x] PR link added to JIRA

- [x] Tested by QA
- [ ] PO's approval

Screenshot 2019-07-08 at 14 09 50

### DIFF
--- a/src/modules/tests/tests.selector.ts
+++ b/src/modules/tests/tests.selector.ts
@@ -73,4 +73,5 @@ export const getActivityCodeBySlotId = (testsModel: TestsModel, id: number): Act
   if (testsModel && testsModel.startedTests && testsModel.startedTests[id]) {
     return testsModel.startedTests[id].activityCode;
   }
+  return null;
 };
--- a/src/modules/tests/tests.selector.ts
+++ b/src/modules/tests/tests.selector.ts
@@ -1,5 +1,5 @@
 import { TestStatus } from './test-status/test-status.model';
-import { StandardCarTestCATBSchema, JournalData } from '@dvsa/mes-test-schema/categories/B';
+import { StandardCarTestCATBSchema, JournalData, ActivityCode } from '@dvsa/mes-test-schema/categories/B';
 import { TestsModel } from './tests.model';
 import { activityCodeModelList } from '../../pages/office/components/activity-code/activity-code.constants';
 import { testReportPracticeSlotId, end2endPracticeSlotId } from '../../shared/mocks/test-slot-ids.mock';
@@ -67,4 +67,10 @@ export const isEndToEndPracticeTest = (tests: TestsModel) : boolean =>
 
 export const getAllTestStatuses = (test: TestsModel): { [slotId: string]: TestStatus; } => {
   return test.testStatus;
+};
+
+export const getActivityCodeBySlotId = (testsModel: TestsModel, id: number): ActivityCode => {
+  if (testsModel && testsModel.startedTests && testsModel.startedTests[id]) {
+    return testsModel.startedTests[id].activityCode;
+  }
 };
--- a/src/pages/journal/components/test-outcome/__tests__/test-outcome.spec.ts
+++ b/src/pages/journal/components/test-outcome/__tests__/test-outcome.spec.ts
@@ -11,6 +11,7 @@ import { StartTest, ActivateTest } from '../../../journal.actions';
 import { TestStatus } from '../../../../../modules/tests/test-status/test-status.model';
 import { OFFICE_PAGE, COMMUNICATION_PAGE } from '../../../../page-names.constants';
 import { DateTime, Duration } from '../../../../../shared/helpers/date-time';
+import { SlotDetail } from '../../../../../shared/models/DJournal';
 
 describe('Test Outcome', () => {
   let fixture: ComponentFixture<TestOutcomeComponent>;
@@ -18,10 +19,10 @@ describe('Test Outcome', () => {
   let store$: Store<StoreModel>;
   let navController: NavController;
 
-  const testSlotDetail = {
+  const testSlotDetail: SlotDetail = {
     duration: 57,
     slotId: 123,
-    start: new DateTime(),
+    start: new DateTime().toString(),
   };
 
   beforeEach(async(() => {
@@ -185,7 +186,7 @@ describe('Test Outcome', () => {
         const dateTime = new DateTime();
         dateTime.subtract(1, Duration.DAY);
 
-        component.slotDetail.start = dateTime;
+        component.slotDetail.start = dateTime.toString();
         component.testStatus = TestStatus.Booked;
         fixture.detectChanges();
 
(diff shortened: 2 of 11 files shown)


ammarhaiderbjss:
Quick screenshot?

src/pages/journal/components/test-outcome/test-outcome.ts
```diff
@@ -10,6 +10,17 @@ import { end2endPracticeSlotId } from '../../../../shared/mocks/test-slot-ids.mo
 import { COMMUNICATION_PAGE, OFFICE_PAGE, PASS_FINALISATION_PAGE } from '../../../page-names.constants';
 import { ModalEvent } from '../../journal-rekey-modal/journal-rekey-modal.constants';
 import { DateTime, Duration } from '../../../../shared/helpers/date-time';
+import { SlotDetail } from '../../../../shared/models/DJournal';
```
ammarhaiderbjss: How come we're not using the centrailsed schema from the @dvsa npm org?

src/pages/journal/components/test-outcome/test-outcome.ts
```diff
@@ -10,6 +10,17 @@ import { end2endPracticeSlotId } from '../../../../shared/mocks/test-slot-ids.mo
...
+import { Observable } from 'rxjs/Observable';
+import { Subscription } from 'rxjs/Subscription';
+import { ActivityCode } from '@dvsa/mes-test-schema/categories/B';
+import { getTests } from '../../../../modules/tests/tests.reducer';
+import { getActivityCodeBySlotId } from '../../../../modules/tests/tests.selector';
+import { map } from 'rxjs/operators';
+
+interface TestOutcomeComponentState {
```
Cavalier07: I believe sub components should be as dumb as possible and should be connected to the store as a last resort. 

Can the slot component itself get the data and pass it in. I don't think any of the sub components of the test slot are connected to the store.

src/pages/journal/components/test-outcome/__tests__/test-outcome.spec.ts
```diff
@@ -223,9 +220,9 @@ describe('Test Outcome', () => {
       it('should hide rekey button for a resumed test if date is today', () => {
         component.slotDetail = testSlotDetail;
 
-        const dateTime = new Date();
+        const dateTime = new DateTime();
```
Cavalier07: Nice :) Moving from JS Date to our helper

src/modules/tests/tests.selector.ts
```diff
@@ -68,3 +68,9 @@ export const isEndToEndPracticeTest = (tests: TestsModel) : boolean =>
 export const getAllTestStatuses = (test: TestsModel): { [slotId: string]: TestStatus; } => {
   return test.testStatus;
 };
+
+export const getActivityCodeBySlotId = (testsModel: TestsModel, id: number): ActivityCode => {
+  if (testsModel && testsModel.startedTests && testsModel.startedTests[id]) {
```
ammarhaiderbjss: What happens here in the case where one of these conditions is false?

src/modules/tests/tests.selector.ts
```diff
@@ -68,3 +68,9 @@ export const isEndToEndPracticeTest = (tests: TestsModel) : boolean =>
...
+    return testsModel.startedTests[id].activityCode;
+  }
+};
```
rossfauldsbjss: I think it would be better to explicitly return null rather than not returning anything and relying on undefined.

src/pages/journal/components/test-outcome/test-outcome.ts
rossfauldsbjss: Second this - it needs to use the @dvsa npm org.  I think we should remove the Djournal building, so that code completion doesn't reference this first, and should put a stop to this going forward.

Cavalier07:
Looking at the screenshot shouldn't the activity code be more centred not aligned to the right as it is at the moment

src/pages/journal/components/test-outcome/test-outcome.ts
Cavalier07: @ammarhaiderbjss  - another thing to go on that ticket ;)

roberthall12: Yep - was auto imported so have now fixed

src/modules/tests/tests.selector.ts
roberthall12: Have changed to return `null` if doesn't meet these conditions

src/modules/tests/tests.selector.ts
roberthall12: Have changed to return null if doesn't meet these conditions

src/pages/journal/components/test-outcome/test-outcome.ts
roberthall12: Okay only did it this way as it was easier to implement but will change

roberthall12:
> Looking at the screenshot shouldn't the activity code be more centred not aligned to the right as it is at the moment

Oli is happy with the alignment, the size was too small so have changed it from `h4` to `h2`

src/pages/journal/components/test-slot/test-slot.ts
```diff
@@ -54,11 +58,15 @@ export class TestSlotComponent implements SlotComponent, OnInit, OnDestroy {
...
+        select(getTests),
+        map(tests => getActivityCodeBySlotId(tests, this.slot.slotDetail.slotId)),
+      ),
     };
 
-    const { testStatus$ } = this.componentState;
+    const { testStatus$, testActivityCode$ } = this.componentState;
 
-    this.subscription = merge(testStatus$).subscribe();
+    this.subscription = merge(testStatus$, testActivityCode$).subscribe();
```
ammarhaiderbjss: Do we need this `.subscribe` if we are using the `| async` pipe in the template?

roberthall12: Yep - hadn't realised they provide the same functionality so have removed

src/pages/journal/components/test-slot/test-slot.ts
```diff
@@ -54,10 +58,13 @@ export class TestSlotComponent implements SlotComponent, OnInit, OnDestroy {
...
       ),
+      testActivityCode$: this.store$.pipe(
+        select(getTests),
+        map(tests => getActivityCodeBySlotId(tests, this.slot.slotDetail.slotId)),
+      ),
     };
 
     const { testStatus$ } = this.componentState;
-
     this.subscription = merge(testStatus$).subscribe();
```
ammarhaiderbjss: @roberthall12 We can also get rid of this subscription and therefore the call to unsubscribe too. Realise its not part of your PR but would be nice to keep things clean. Happy with this once that's done so will approve 👍

PR #567, opened 2019-07-12
Mes 2931 serious dangerous faults analytics

buzzub:
- Add effects for serious and dangerous faults
- Ensure the analytics effects return a type because as I now understand, returning `of({})` will produce an error `actions must have a type property`... and this breaks the test report 😞  So they now return an action of `AnalyticRecorded or `AnalyticNotRecorded` instead - which is effectively a dummy / void action. But this does mean that tests can now check to see the correct action was returned.

A future PR will need to look at Logs and other Analytics to fix similar issues.

## Pull Request checklist:

- [x] commits are squashed

- [x] Tests are passing (AND WORKS ON DEVICE!)
- [x] PR link added to JIRA

- [ ] Tested by QA
- [ ] PO's approval

### DIFF
--- a/src/pages/test-report/test-report.analytics.effects.ts
+++ b/src/pages/test-report/test-report.analytics.effects.ts
@@ -11,7 +11,7 @@ import {
 import * as testReportActions from '../../pages/test-report/test-report.actions';
 import * as testDataActions from '../../modules/tests/test-data/test-data.actions';
 import { getTests } from '../../modules/tests/tests.reducer';
-import { isTestReportPracticeTest } from '../../modules/tests/tests.selector';
+import { isPracticeMode } from '../../modules/tests/tests.selector';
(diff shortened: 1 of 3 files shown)


src/pages/test-report/test-report.analytics.effects.ts
```diff
@@ -39,10 +37,73 @@ export class TestReportAnalyticsEffects {
...
+    }),
+  );
+
+  @Effect()
+  toggleDangerousFaultMode$ = this.actions$.pipe(
+    ofType(testReportActions.TOGGLE_DANGEROUS_FAULT_MODE),
+    withLatestFrom(
+      this.store$.pipe(
+        select(getTests),
+        map(isTestReportPracticeTest),
```
Cavalier07: If your trying to avoid practice mode this selector on catches the Test Report one not the End To End . You need to use this selector instead - isPracticeMode

PR #569, opened 2019-07-15
Adding role to config

MarkPatt89:
- Adding examiner role to config for future stories to conditionally show different journeys
## Pull Request checklist:

- [x] commits are squashed

- [x] Tests are passing
- [x] PR link added to JIRA

- [ ] Tested by QA
- [ ] PO's approval

### DIFF
--- a/src/providers/app-config/app-config.model.ts
+++ b/src/providers/app-config/app-config.model.ts
@@ -1,3 +1,5 @@
+import { ExaminerRole } from './constants/examiner-role.constants';
+
 export type AppConfig = {
   configUrl: string
   googleAnalyticsId: string,
@@ -17,7 +19,7 @@ export type AppConfig = {
   },
   approvedDeviceIdentifiers: string[],
   timeTravelDate?: string,
-  role: string;
+  role: ExaminerRole;
   journal: {
     journalUrl: string,
     autoRefreshInterval: number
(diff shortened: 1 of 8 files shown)


src/providers/app-config/app-config.model.ts
```diff
@@ -17,6 +17,7 @@ export type AppConfig = {
   },
   approvedDeviceIdentifiers: string[],
   timeTravelDate?: string,
+  role: string;
```
Cavalier07: Same as other could we have the roles in an enum DE and LDTM

MarkPatt89: I'm not sure that it is worth it, as the value being provided will be controlled by what's in Dynamo. Also, we are not limited to DE and LDTM. This may grow in the future.

MarkPatt89: For further context, we aren't storing a boolean in Dynamo, it's a string representation of the role.

Cavalier07: It just makes the code a bit safer and is easier to reference/use it throughout the app. I can see the person who picks up the ticket and has to limit components by roles preferring an enum over strings.

Cause it may change in the future is not really a valid reason as everything can change in the future so we'd do no work ;)

rstepnowski: Assuming that there may be other roles in the system (in the future), having an enum already in place would be helpful. Ans we don't need to worry about case-sensitivity when checking the role

PR #574, opened 2019-07-16
MES-2920: Performance Issue Search Results

szabolcs-keresztes:
Performance issue when loading more than 30 test results has been solved.

### DIFF
--- a/src/pages/test-results-search/test-results-search.scss
+++ b/src/pages/test-results-search/test-results-search.scss
@@ -66,4 +66,36 @@ page-test-results-search {
     margin-right: 0;
     width: 100%;
   }
+
+  .search-inputs {
+    position: absolute;
+    width: 800px;
+    z-index: 100;
+  }
+
+  @media screen and (min-width: 900px) {
+    .search-inputs {
+      position: absolute;
+      width: 1080px;
+      z-index: 100;
+    }
+  }
+
+  .test-results-list {
+
+    .list-header {
+      height: 480px;
+      visibility: hidden;
+    }
+
+    .item-ios {
+      background-color: map-get($colors, 'mes-background');
+      padding: 0;
+    }
+
+    ion-label.label.label-ios {
+      margin: 0;
+    }
+  }
+  
 }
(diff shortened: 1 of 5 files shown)


src/pages/test-results-search/test-results-search.scss
```diff
@@ -66,4 +66,36 @@ page-test-results-search {
...
+    position: absolute;
+    width: 800px;
+    z-index: 100;
+  }
+
+  @media screen and (min-width: 900px) {
+    .search-inputs {
+      position: absolute;
+      width: 1080px;
+      z-index: 100;
```
ammarhaiderbjss: Did we need still this z-index then?

src/pages/test-results-search/test-results-search.ts
```diff
@@ -101,4 +102,11 @@ export class TestResultsSearchPage extends BasePageComponent {
       .subscribe();
   }
 
+  myHeaderFn(record: any, recrodIndex: any): string {
```
ammarhaiderbjss: typo here

src/pages/test-results-search/test-results-search.scss
ammarhaiderbjss: ignore me, thought this was .list-header

src/pages/test-results-search/test-results-search.ts
ammarhaiderbjss: What's this method for btw?

src/pages/test-results-search/test-results-search.scss
szabolcs-keresztes: The z-index would bring the search tabs in from of the the ion-list header.

src/pages/test-results-search/test-results-search.ts
szabolcs-keresztes: The `ion-list` needs some function to determine which items in the list need header. In our case just the first item will need header, but we purely use this only for design so we don't need to return any information just an empty string.

Corrected the typo.

PR #576, opened 2019-07-16
Update test report analytics effects and tests to cater for practice mode

buzzub:
This is the first of many PRs for changes to analytics, all necessary to split out practice analytics from regular analytics during practice tests and end to end practice.

- Created a helper function to prefix analytics events with "practice mode", "practice test" or nothing depending on the test state.
- Changed the analytics effects so that practice events are logged, but now prefixed e.g. `category: "practice test - test report"` `event: "practice test - add driving fault"` `label: "Junctions - Approach speed"`.
- Updated the tests so that they are much better. Done is passed into the test and they now also fail if the effect does not emit a result.

## Pull Request checklist:

- [ ] commits are squashed

- [x] Tests are passing
- [x] PR link added to JIRA

- [ ] Tested by QA
- [ ] PO's approval

### DIFF
--- a/src/providers/analytics/analytics.model.ts
+++ b/src/providers/analytics/analytics.model.ts
@@ -14,7 +14,7 @@ export interface IAnalyticsProvider {
 }
 
 export enum AnalyticsScreenNames {
-  CONTACT_DETAILS = 'contact details screen', // this may need removing as could be candidate details now
+  COMMUNICATION = 'communication screen',
   HEALTH_DECLARATION = 'health declaration screen',
   JOURNAL = 'journal screen',
   PASS_TEST_SUMMARY = 'pass test summary screen',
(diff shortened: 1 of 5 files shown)


src/providers/analytics/analytics.model.ts
```diff
@@ -42,6 +42,8 @@ export enum AnalyticsEventCategories {
   POST_TEST = 'post-test',
   TEST_REPORT = 'test report',
   TERMINATION = 'test termination',
+  PRACTICE_TEST = 'practice test',
```
Cavalier07: If I was being picky practice test is not a screen name ;)

buzzub: As you now know, it's actually under `AnalyticsEventCategories` ;)

PR #583, opened 2019-07-18
MES-2948: fix pass cert length

MarkPatt89:
- update `pass-finalisation` validation to enforce a max length of 8 chars for `pass-certificate-numbers`

Screenshot 2019-07-19 at 12 09 32
Screenshot 2019-07-19 at 12 09 43

## Pull Request checklist:

- [x] commits are squashed

- [x] Tests are passing
- [ ] PR link added to JIRA

- [ ] Tested by QA
- [ ] PO's approval

### DIFF
--- a/src/pages/pass-finalisation/__tests__/pass-finalisation.spec.ts
+++ b/src/pages/pass-finalisation/__tests__/pass-finalisation.spec.ts
@@ -44,17 +44,39 @@ describe('PassFinalisationPage', () => {
       });
   }));
 
-  describe('Class', () => {
-    describe('onSubmit', () => {
-      // Unit tests for the components TypeScript class
-      it('should dispatch the PersistTests action', () => {
-        const form = component.form;
-        form.get('provisionalLicenseProvidedCtrl').setValue(true);
-        form.get('passCertificateNumberCtrl').setValue(true);
-        form.get('transmissionCtrl').setValue('Manual');
-        component.onSubmit();
-        expect(store$.dispatch).toHaveBeenCalledWith(new PersistTests());
-      });
+  describe('onSubmit', () => {
+    // Unit tests for the components TypeScript class
+    it('should dispatch the PersistTests action', () => {
+      const form = component.form;
+      form.get('provisionalLicenseProvidedCtrl').setValue(true);
+      form.get('passCertificateNumberCtrl').setValue('A123456*');
+      form.get('transmissionCtrl').setValue('Manual');
+      component.onSubmit();
+      expect(store$.dispatch).toHaveBeenCalledWith(new PersistTests());
+    });
+  });
+  describe('formControls', () => {
+    it('should contain a maxlength validation error when passCertificateNumberCtrl fails to meet maxlength', () => {
+      const formCtrl = component.form.controls['passCertificateNumberCtrl'];
+      formCtrl.setValue('A123456B1');
+      expect(formCtrl.hasError('maxlength')).toBe(true);
+    });
+    it('should contain no validation errors when passCertificateNumberCtrl ends with digit', () => {
+      const formCtrl = component.form.controls['passCertificateNumberCtrl'];
+      formCtrl.setValue('A1234567');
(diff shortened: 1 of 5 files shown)


src/pages/pass-finalisation/__tests__/pass-finalisation.spec.ts
```diff
@@ -59,10 +59,34 @@ describe('PassFinalisationPage', () => {
...
       const form = component.form;
       form.get('provisionalLicenseProvidedCtrl').setValue(true);
-      form.get('passCertificateNumberCtrl').setValue(true);
+      form.get('passCertificateNumberCtrl').setValue('A123456*');
       form.get('transmissionCtrl').setValue('Manual');
       component.onSubmit();
       expect(store$.dispatch).toHaveBeenCalledWith(new PersistTests());
     });
   });
+  fdescribe('formControls', () => {
```
johnmorleybjss: Focused test suite needs removing.

MarkPatt89: ahh, thanks John.

src/pages/pass-finalisation/pass-finalisation.ts
```diff
@@ -187,16 +187,26 @@ export class PassFinalisationPage extends PracticeableBasePageComponent {
   }
 
   getFormValidation(): { [key: string]: FormControl } {
+    const passCerticateFormat = new RegExp(/\D{1}\d{6}\W{1}/);
```
johnmorleybjss: Won't this accept symbols for the first character (anything that's not a digit)? Shouldn't you use `[A-Z]`?

src/pages/pass-finalisation/__tests__/pass-finalisation.spec.ts
MarkPatt89: Done

src/pages/pass-finalisation/pass-finalisation.ts
MarkPatt89: You're 100% right, fixed and pushing changes once I check out the comment above.

MarkPatt89: Spoke with Andy around the last digit, this has also been amended.

buzzub:
Shouldn't the error message say what the allowed / expected format is? E.g. https://design-system.service.gov.uk/components/error-summary/ shows "Enter a postcode, like AA1 1AA"

MarkPatt89:
> Shouldn't the error message say what the allowed / expected format is? E.g. https://design-system.service.gov.uk/components/error-summary/ shows "Enter a postcode, like AA1 1AA"

I had queried this, they will have the number on the certificate in front of them. Also verified the content with Oli.

buzzub:
I know that the examiner will have the certificate in front of them (and this error is unlikely to appear), but I feel that we should let them know if they should be prefixing with the letter or not, especially if we're asking for a "**pass certificate number**".

I think we either need to include an example in the field label or an explanation in the error text. As it stands, the error does not help the user resolve it.

MarkPatt89:
> I know that the examiner will have the certificate in front of them (and this error is unlikely to appear), but I feel that we should let them know if they should be prefixing with the letter or not, especially if we're asking for a "**pass certificate number**".
> 
> I think we either need to include an example in the field label or an explanation in the error text. As it stands, the error does not help the user resolve it.

I'm happy to change it, I'd rather the design guys approval first. I'll query it with them.

buzzub:
> I'm happy to change it, I'd rather the design guys approval first. I'll query it with them.

Just cite this from the GDS guide: 

**> Above all, aim for clarity.**

MarkPatt89:
> Looks good Mark - do we need to do anything to enforce this on the test result service or via the schema

No, we don't need to concern ourselves with that. Catching this at the application layer will suffice. The issue was spotted by Chris, as RSIS MI DB was rejecting test results due to them not meeting the desired pattern.

PR #584, opened 2019-07-18
MES-2939: Limiting text areas for 1000 characters only

szabolcs-keresztes:
- Limiting Candidate description (1000 chars max)
- Limiting all dangerous, serious, driving faults (1000 chars max)
- Writing unit tests for these

### DIFF
--- a/src/pages/office/office.ts
+++ b/src/pages/office/office.ts
@@ -607,9 +607,6 @@ export class OfficePage extends PracticeableBasePageComponent {
   }
 
   additionalInformationChanged(additionalInformation: string): void {
-    if (additionalInformation.length > 5000) {
-      return;
-    }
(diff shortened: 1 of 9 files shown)


src/pages/office/office.ts
```diff
@@ -596,6 +596,9 @@ export class OfficePage extends PracticeableBasePageComponent {
   }
 
   candidateDescriptionChanged(candidateDescription: string) {
+    if (candidateDescription.length > 1000) {
```
johnmorleybjss: Perhaps it would be better to prevent the presentational components from emitting when they are invalid instead of having to guard in the container component?

szabolcs-keresztes: Good idea, will make the change

PR #605, opened 2019-07-25
Mes 3055 outcome changed event label

szabolcs-keresztes:
Adding a label to outcome changed analytics event

### DIFF
--- a/src/pages/test-report/test-report.effects.ts
+++ b/src/pages/test-report/test-report.effects.ts
@@ -105,7 +105,9 @@ export class TestReportEffects {
 
             const actions: Action[] = [new testsActions.SetActivityCode(result)];
             if (!isEmpty(currentTest.activityCode) && currentTest.activityCode !== result) {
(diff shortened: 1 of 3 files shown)


src/pages/test-report/test-report.effects.ts
```diff
@@ -105,7 +105,8 @@ export class TestReportEffects {
 
             const actions: Action[] = [new testsActions.SetActivityCode(result)];
             if (!isEmpty(currentTest.activityCode) && currentTest.activityCode !== result) {
-              actions.push(new testsActions.TestOutcomeChanged());
+              const label = result === '1' ? 'fail to pass' : 'pass to fail';
```
johnmorleybjss: Any requirement to know when terminate was changed?

szabolcs-keresztes: That's what I was thinking about as well. But we are purely interested here whether it changed from fail to pass or pass to fail. Would be used to prevent fraud

PR #624, opened 2019-08-01
Bug mes 2923 fix transport security in plist

rossfauldsbjss:
Our mobile app build is accidentally disabling ATS, meaning the app can issue HTTP requests and isn't forced to only use HTTPS. This was found in the ITHC, but also violates the MES HLD which promised we would use ATS.

This fix modifies the plist so that ATS is enabled (NSAllowsArbitraryLoads is false), it does so by adding an npm package, a script  and a hook to call the script.


chrisnappin (approved):
Looks good

roberthall12 (approved):

ghost (approved):

MarkPatt89 (approved):

PR #625, opened 2019-08-01
Mes 2929 - pass finalisation move fields

roberthall12:
Moved fields before green screen for auto saving
- Moved `d255, debrief witnessed, language pref` radio buttons to pass finalisation page.
- Fixed validation of the radio buttons

## Pull Request checklist:

- [ ] branch rebased against the latest develop
- [ ] commits are squashed

- [ ] Tests are passing
- [ ] PR link added to JIRA

- [ ] Tested by QA
- [ ] PO's approval

### DIFF
--- a/src/components/components.module.ts
+++ b/src/components/components.module.ts
@@ -18,6 +18,9 @@ import { TickIndicatorComponent } from './tick-indicator/tick-indicator';
 import { ModalAlertTitleComponent } from './modal-alert-title/modal-alert-title';
 import { ModalReturnButtonComponent } from './modal-return-button/modal-return-button';
 import { ErrorMessageComponent } from './error-message/error-message';
+import { D255Component } from '../pages/office/components/d255/d255';
+import { DebriefWitnessedComponent } from '../pages/office/components/debrief-witnessed/debrief-witnessed';
+import { LanguagePreferencesComponent } from '../pages/office/components/language-preference/language-preferences';
 
 @NgModule({
   declarations: [
@@ -35,6 +38,9 @@ import { ErrorMessageComponent } from './error-message/error-message';
     ModalAlertTitleComponent,
     ModalReturnButtonComponent,
     ErrorMessageComponent,
+    LanguagePreferencesComponent,
+    D255Component,
+    DebriefWitnessedComponent,
   ],
   imports: [
     SignaturePadModule,
@@ -58,6 +64,9 @@ import { ErrorMessageComponent } from './error-message/error-message';
     ModalAlertTitleComponent,
     ModalReturnButtonComponent,
     ErrorMessageComponent,
+    LanguagePreferencesComponent,
+    D255Component,
+    DebriefWitnessedComponent,
   ],
 })
 export class ComponentsModule { }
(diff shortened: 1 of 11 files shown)


src/pages/pass-finalisation/pass-finalisation.ts
```diff
@@ -68,9 +84,8 @@ export class PassFinalisationPage extends PracticeableBasePageComponent {
   passCertificateCtrl: string = 'passCertificateNumberCtrl';
   @ViewChild('passCertificateNumberInput')
   passCertificateNumberInput: ElementRef;
-
   inputSubscriptions: Subscription[] = [];
-
+  testOutcome: string = '1';
```
Cavalier07: can't see a use for this

roberthall12: The radio buttons being displayed and validation from the office page are driven by the outcome.

src/pages/office/components/d255/d255.html
```diff
@@ -1,7 +1,7 @@
 <ion-row *ngIf="display" class="mes-validated-row mes-row-separator" id="d255-card" [formGroup]="formGroup">
   <div class="validation-bar" [class.ng-invalid]="invalid">
   </div>
-  <ion-col class="component-label" col-32 align-self-center>
+  <ion-col class="component-label" col-36 align-self-center>
```
TomBillingtonUK, line 4: These components need moving to the pass finalisation components folder as the end result is to remove them from the office.

src/components/components.module.ts
```diff
@@ -18,6 +18,9 @@ import { TickIndicatorComponent } from './tick-indicator/tick-indicator';
 import { ModalAlertTitleComponent } from './modal-alert-title/modal-alert-title';
 import { ModalReturnButtonComponent } from './modal-return-button/modal-return-button';
 import { ErrorMessageComponent } from './error-message/error-message';
+import { D255Component } from '../pages/office/components/d255/d255';
```
TomBillingtonUK, line 21: These all need removing when you do the next bit (remove the fields from the office page) as they won't technically be shared.

src/pages/pass-finalisation/pass-finalisation.ts
rstepnowski: There should be at least an enum somewhere for those test outcomes/activity codes - right?

TomBillingtonUK: This feels a bit hard coded , why do we need this and why do we need to pass it into the debrief witnessed field.

If it is needed can we change it to ActivityCodes.Pass instead of using a string.

TomBillingtonUK (approved):

src/pages/pass-finalisation/pass-finalisation.ts
roberthall12: It is just used to drive the visibility of the field because this will be reused on each of the pass, fail journeys and each termination code. I've changed it to `ActivityCodes.Pass`

src/components/components.module.ts
roberthall12: They will be shared between the `pass-finalisation` page and the new page introduced before the green screen.

TomBillingtonUK: Ok , can we move the components into the root components folder then

ghost (approved):

PR #627, opened 2019-08-01
Feature - character count on input fields

buzzub:
This is an attempt at creating a method for displaying character counts below input fields.

I created a directive to calculate the number of characters remaining and followed the [GDS guidelines](https://design-system.service.gov.uk/components/character-count/)

The character count is added manually to allow for flexibility and is separate from any existing maxLength validation. It reuses similar styles to the existing validation errors.

Some screenshots below:

IMG_0080

IMG_0082

IMG_0079

IMG_0078

### DIFF
--- a/src/pages/office/components/fault-comment/fault-comment.ts
+++ b/src/pages/office/components/fault-comment/fault-comment.ts
@@ -40,6 +40,7 @@ export class FaultCommentComponent implements OnChanges {
   @Output()
   faultCommentChange = new EventEmitter<CommentedCompetency>();
 
+  faultCommentCharsRemaining: number = null;
   static readonly fieldName: string = 'faultComment';
   constructor(private outcomeBehaviourProvider: OutcomeBehaviourMapProvider) { }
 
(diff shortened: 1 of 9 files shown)


szabikr (approved):

MarkPatt89 (approved):

src/pages/office/components/fault-comment/fault-comment.ts
```diff
@@ -81,6 +79,20 @@ export class FaultCommentComponent implements OnChanges {
     this.faultCommentChange.emit(commentedCompetency);
   }
 
+  characterCountChanged(charactersRemaining: number) {
+    this.faultCommentCharsRemaining = charactersRemaining;
+  }
+
+  getCharacterCountText() {
```
TomBillingtonUK, line 86: Feels like these should be helpers as they are in multiple classes

TomBillingtonUK (approved):
Looks good

Should this be on the additional info (optional) field as well

src/pages/office/components/fault-comment/fault-comment.ts
szabikr: It could be, but each component has the option to set individual parameters. So that helper function would have a long list of parameters which wouldn't really solve anything. 

And we used this only 2 places so far. The rule is, when you're about the implement the same thing for the 4th time you could start looking into the generalisation of the solution

szabikr:
Additional info can have _indefinite_ number of characters

PR #639, opened 2019-08-06
Mes 3020 search tests to rekey

szabikr:
Creating RekeySearchEffect which will deal with the RekeySearchProvider to get the booked test slot from the lambda. Eventually store the testSlot data to the store and reselect to the RekeySearchPage.


src/app/app.module.ts
```diff
@@ -103,6 +104,7 @@ const enableDevTools = environment && environment.enableDevTools;
     { provide: HAMMER_GESTURE_CONFIG, useClass: IonicGestureConfig },
     OutcomeBehaviourMapProvider,
     LogHelper,
+    RekeySearchProvider,
```
TomBillingtonUK: Does it need to be a global provider can it not be in the RekeySearch Page Module

TomBillingtonUK (approved):
Nice Work

src/app/app.module.ts
szabikr: Moved the declaration to the RekeySearchPage module.

src/providers/rekey-search/rekey-search.ts
```diff
@@ -0,0 +1,26 @@
...
+
+@Injectable()
+export class RekeySearchProvider {
+
+  constructor(
+    private httpClient: HttpClient,
+    private urlProvider: UrlProvider,
+  ) { }
+
+  getTest(rekeySearchParams: RekeySearchParams): Observable<any> {
```
roberthall12: Can this be a better type? `Observable<TestSlot>`?

src/pages/rekey-search/rekey-search.ts
```diff
@@ -1,19 +1,33 @@
...
   templateUrl: 'rekey-search.html',
 })
-export class RekeySearchPage extends BasePageComponent {
+export class RekeySearchPage extends BasePageComponent implements OnInit {
+
+  pageState: RekeySearchPageState;
 
+  staffNumber: string = '';
+  applicationReference: string = '';
+  searchResults: any[] = [];
```
roberthall12: Same again, could we use a more safe type?

roberthall12 (approved):
Looks good - would like to avoid using `any` type if possible

src/providers/rekey-search/rekey-search.ts
szabikr: Added `TestSlot` type

src/pages/rekey-search/rekey-search.ts
szabikr: Added `TestSlot` type

PR #654, opened 2019-08-07
MES-2689: Dispatch CandidateDetailsSeen on view enter

jameshighfield:
We were dispatching an action when state updates and that action updates the same state so there’s an infinite loop of actions being dispatched. Now dispatching the action on view enter.

Possible fix for MES-2689.


ammarhaiderbjss (approved):

src/pages/journal/journal.reducer.ts
```diff
@@ -23,6 +23,9 @@ export function journalReducer(state = initialState, action: journalActions.Jour
         error: { message: '', status: 0, statusText: '' },
       };
     case journalActions.CANDIDATE_DETAILS_SEEN:
+      if (state.checkComplete.findIndex(checkComplete => checkComplete.slotId === action.slotId) !== -1) {
```
szabikr, line 26: Just want to mention that I think there is too much logic in the reducer with this. No change required, only if you prefer to do

szabikr (approved):

roberthall12 (approved):

PR #688, opened 2019-08-15
MES-3021: Display Candidate Details for test slot on rekey process 

szabikr:
The Candidate Details Page no longer queries the store but gets the slot data as a parameter.

### DIFF
--- a/src/components/test-slot/candidate-link/candidate-link.ts
+++ b/src/components/test-slot/candidate-link/candidate-link.ts
@@ -12,7 +12,7 @@ import { CANDIDATE_DETAILS_PAGE, FAKE_CANDIDATE_DETAILS_PAGE } from '../../../pa
 })
 export class CandidateLinkComponent {
   @Input()
-  slotId: number | string;
+  slot: any;
 
   @Input()
   slotChanged: boolean;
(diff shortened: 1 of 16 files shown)


src/components/test-slot/candidate-link/candidate-link.ts
```diff
@@ -12,7 +12,7 @@ import { CANDIDATE_DETAILS_PAGE, FAKE_CANDIDATE_DETAILS_PAGE } from '../../../pa
 })
 export class CandidateLinkComponent {
   @Input()
-  slotId: number | string;
+  slot: any;
```
rossfauldsbjss, line 15: can we strongly type this instead of any?

rossfauldsbjss (approved):
looks good

src/components/test-slot/candidate-link/candidate-link.ts
szabikr: We could but then we have to delete a unit test because the `slotId` is a string for practice test and in the schema is a number.

TomBillingtonUK (approved):

src/components/test-slot/candidate-link/candidate-link.ts
szabikr: I'll resolve this conversation and merge the PR. But we can have a chat about this and come up with a way of getting that strongly typed.

PR #703, opened 2019-08-20
re-enable subscription when coming back from terminate page

rossfauldsbjss:
When legal requirements aren't fulfilled and the test is terminated, if you go back in to the test report the legal requirements aren't acknowledged as being completed when they have been.

The solution was to re-enable the subscription so that changes were registered in state.


MarkPatt89 (approved):
Cool, is it possible to get a screenshot for this?

szabikr (approved):

PR #735, opened 2019-09-04
MES-3526: Moving checkComplete flags to slotItem

szabikr:
Initially the JournalPage was the first page that loaded in the mobile app. And that module brought its slice of the state and all that. After introducing the Dashboard page we don't need to go to the Journal in order to search for a test to rekey from the RekeySearchPage.
The fact that we are reusing the the same test journey for normal test or rekey from both pages (Journal, RekeySearch) requires us to be generic. So the checkComplete array had to be removed and the information stored there had to be moved to the individual SlotItems.

### DIFF
--- a/mock/generate-local-journal.ts
+++ b/mock/generate-local-journal.ts
@@ -17,6 +17,18 @@ const localJournal: ExaminerWorkSchedule = {
     individualId: 9000000,
   },
   nonTestActivities: [],
+  personalCommitments: [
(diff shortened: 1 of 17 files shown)


buzzub (approved):

MarkPatt89 (approved):

amobarak (approved):
Superb, great job and works nicely!

PR #757, opened 2019-09-12
MES-3574 reload app config when app resumes

buzzub:
## Description
When the app starts up and loads the configuration successfully, store the current date in the `appInfo` part of the store.

Then using the platform 'resume` event, check if the date the app config was loaded is different to today's date, and if so, reload the app by setting the app root as the login page.

## Checklist

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -37,7 +37,7 @@ export class App {
         this.loadAppInfo();
         if (this.platform.is('ios')) {
           this.configureAccessibility();
-          this.createPlatformSubscriptions();
+          this.configurePlatformSubscriptions();
         }
       });
   }
@@ -65,7 +65,7 @@ export class App {
     this.store$.dispatch(new LoadAppInfo());
   }
 
-  createPlatformSubscriptions() {
+  configurePlatformSubscriptions() {
     const merged$ = merge(
       this.platform.resume.pipe(map(this.onAppResumed)),
       this.platform.pause.pipe(map(this.onAppSuspended)),
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -21,6 +21,7 @@ export class App {
   textZoom: number = 100;
   increasedContrast: Boolean = false;
 
+  private platformSubscription: Subscription;
   private subscription: Subscription;
 
   constructor(
@@ -33,13 +34,18 @@ export class App {
       .then(() => {
         this.configureLocale();
         this.configureStatusBar();
-        this.configureAccessibility();
         this.loadAppInfo();
-        this.configurePlatformSubscriptions();
+        if (this.platform.is('ios')) {
+          this.configureAccessibility();
+          this.createPlatformSubscriptions();
+        }
       });
   }
 
   ionViewWillUnload() {
+    if (this.platformSubscription) {
+      this.platformSubscription.unsubscribe();
+    }
     if (this.subscription) {
       this.subscription.unsubscribe();
     }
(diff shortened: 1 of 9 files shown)


src/modules/app-info/app-info.effects.ts
```diff
@@ -27,4 +36,32 @@ export class AppInfoEffects {
...
+    ofType(appInfoActions.APP_RESUMED),
+    concatMap(action => of(action).pipe(
+      withLatestFrom(
+        this.store$.pipe(
+          select(getAppInfoState),
+          select(getDateConfigLoaded),
+        ),
+      ),
+    )),
+    filter(([action, dateConfigLoaded]) => dateConfigLoaded !== this.dateTimeProvider.now().format('YYYY-MM-DD')),
```
ammarhaiderbjss, line 60: Nice 😍

src/modules/app-info/app-info.effects.ts
```diff
@@ -27,4 +36,32 @@ export class AppInfoEffects {
...
+    switchMap(([action, dateConfigLoaded]) => {
+      console.log('App resumed after being suspended. Config was not loaded today... app will refresh');
+      this.navigationProvider.getNav().setRoot(LOGIN_PAGE);
+      return of(new appInfoActions.RestartApp());
```
ammarhaiderbjss, line 64: Is this action used for anything or is it just for info

TomBillingtonUK (commented):
Nice , clean and simple :)

src/modules/app-info/app-info.effects.ts
buzzub: Yeah, it's just for info. Need to return an observable, so figured I'd return a descriptive action. Could easily just return empty `of()` but the benefit of this is I'll be able to write a test.

There's also an `AppSuspended` action. It get's dispatched, but nothing acts on it (yet).

ammarhaiderbjss (approved):
Implementation looks good, happy once tested on an enrolled device :)

mgoldson (commented):
Just looked at this to try and broaden my understanding of the app - looks nice!
One thing I would like to see is a consistent approach to naming actions. Looks like the established pattern is VERB_NOUN using present tense for the verb. Some of the new actions introduced go with NOUN_VERB using past tense for the verb.

buzzub:
> Just looked at this to try and broaden my understanding of the app - looks nice!
> One thing I would like to see is a consistent approach to naming actions. Looks like the established pattern is VERB_NOUN using present tense for the verb. Some of the new actions introduced go with NOUN_VERB using past tense for the verb.



> Just looked at this to try and broaden my understanding of the app - looks nice!
> One thing I would like to see is a consistent approach to naming actions. Looks like the established pattern is VERB_NOUN using present tense for the verb. Some of the new actions introduced go with NOUN_VERB using past tense for the verb.

Good observation. If this refers to `AppSuspended` and `AppResumed`, then I think in this case the actions are correctly named because they are dispatched when the platform plugin events fire and as such they are past tense.

`ResumeApp` wouldn't make sense because that implies the action will cause an effect, where as it was the cordova plugin that did the resuming and the new action is reacting to that.

`ResumedApp` could work I guess but it sounds wrong.

buzzub:
> Implementation looks good, happy once tested on an enrolled device :)

Andy tested and found no problems.

src/app/app.component.ts
```diff
@@ -47,19 +65,29 @@ export class App {
...
+    window.MobileAccessibility.usePreferredTextZoom(true);
+    window.MobileAccessibility.getTextZoom(this.getTextZoomCallback);
+  }
+
+  onAppSuspended = () => {
+    this.store$.dispatch(new AppSuspended());
+  }
+
+  configureAccessibility = () => {
+    window.MobileAccessibility.updateTextZoom();
```
rossfauldsbjss, line 87: wondering about the removal of the window && window.MobileAccessibility line (51).
Is this because window is guaranteed to exists and this was an unnecessary check?

src/modules/app-info/app-info.effects.ts
```diff
@@ -27,4 +36,32 @@ export class AppInfoEffects {
         );
     }),
   );
+
+  @Effect()
+  loadConfigSuccessEffect$ = this.actions$.pipe(
+    ofType(appInfoActions.LOAD_CONFIG_SUCCESS),
+    switchMap(() => {
+      console.log('Config loaded successfully');
```
rossfauldsbjss, line 44: do we need console.logs?

rossfauldsbjss (approved):
looks good. 
Couple of comments, but nothing that affects approval.

src/app/app.component.ts
buzzub: Exactly. And we can assume `MobileAccessibility` will exist too unless the plugin fails.

src/modules/app-info/app-info.effects.ts
buzzub: I think Ammar wanted this. I can't remember exactly why... but it definitely made sense at the time.

src/app/app.component.ts
buzzub: Also, we shouldn't really need to use `window.MobileAccessibility` as this should work:

```
constructor(mobileAccessibility: MobileAccessibility) {
  this.platform.ready()
    .then(() => mobileAccessibility.getTextZoom())
    .then(e => this.zoomLevel = e);
}
``` 
see more at https://github.com/ionic-team/ionic-native/issues/2138

...but it doesn't. The signature is wrong - it doesn't return a promise, it returns a callback ... but you can't pass it a function. Went down a rabbit hole trying to reference the plugin correctly. In the end, decided to leave the code as is because it works.

TomBillingtonUK: How does this react in the browser , we know the plugin won't be there and that line will error

PR #758, opened 2019-09-13
MES-2617 autosave result

edwardwilson:
## Description
Test submission effect now also looks for test data in the write-up state. Those in the WriteUp state will be sent as a partial save. Once saved the state will be moved to Autosaved.

This PR includes all the needed changes to the effects and state for autosave.

## Checklist

## Screenshots (optional)

### DIFF
--- a/src/modules/tests/tests.actions.ts
+++ b/src/modules/tests/tests.actions.ts
@@ -83,7 +83,7 @@ export class SendCompletedTestsFailure implements Action {
 
 export class SendPartialTestSuccess implements Action {
   readonly type = SEND_PARTIAL_TEST_SUCCESS;
-  constructor(public testId: string) {}
+  constructor(public slotId: string) {}
 }
 
 export class SendPartialTestsFailure implements Action {
--- a/src/modules/tests/tests.actions.ts
+++ b/src/modules/tests/tests.actions.ts
@@ -11,6 +11,9 @@ export const SEND_CURRENT_TEST = '[TestsEffects] Send Current Test';
 export const SEND_CURRENT_TEST_SUCCESS = '[Tests] Send Test Success';
 export const SEND_CURRENT_TEST_FAILURE = '[Tests] Send Test Failure';
 
+export const SEND_PARTIAL_TEST_SUCCESS = '[TestsEffects] Send Partial Tests Success';
+export const SEND_PARTIAL_TESTS_FAILURE = '[TestsEffects] Send Partial Tests Failure';
+
 export const PERSIST_TESTS = '[Tests] Persist';
 export const LOAD_PERSISTED_TESTS = '[Tests] Load persisted';
 export const LOAD_PERSISTED_TESTS_SUCCESS = '[Tests] Load persisted success';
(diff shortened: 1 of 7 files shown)


roberthall12:
Can you add a quick description for what's been added?

src/providers/test-submission/__tests__/test-submission.spec.ts
```diff
@@ -57,12 +58,13 @@ describe('TestSubmissionProvider', () => {
         index: 0,
         slotId: '',
         payload: {},
+        status: TestStatus.Completed,
       }] as TestToSubmit[]).subscribe();
 
-      httpMock.expectOne('https://www.example.com/api/v1/test-result');
+      httpMock.expectOne('https://www.example.com/api/v1/test-result?partial=false');
```
roberthall12: I don't think we need `partial=false` for a full submission, see https://wiki.dvsacloud.uk/display/MES/Autosave+LLD

src/providers/test-submission/test-submission.ts
```diff
@@ -69,5 +81,8 @@ export class TestSubmissionProvider {
     };
     return removeNullFields(data);
   }
-
+  removeFieldsForPartialData = (data: StandardCarTestCATBSchema): Partial<StandardCarTestCATBSchema> => {
+    data.testSummary = null;
```
roberthall12: We can't do this as this will remove `D255`, `debriefWitnessed` from the state

roberthall12 (changes requested):
Good - just need to modify the `removeFieldsForPartialData` function to allow us to keep the fields collected in pass/non-pass finalisation pages

src/modules/tests/tests.actions.ts
```diff
@@ -78,6 +81,15 @@ export class SendCompletedTestsFailure implements Action {
   readonly type = SEND_COMPLETED_TESTS_FAILURE;
 }
 
+export class SendPartialTestSuccess implements Action {
+  readonly type = SEND_PARTIAL_TEST_SUCCESS;
+  constructor(public testId: string) {}
```
buzzub: I think this should be named `slotId` instead. `testId` isn't really used anywhere else

src/modules/tests/tests.effects.ts
```diff
@@ -252,15 +253,30 @@ export class TestsEffects {
...
+                ? new testActions.SendPartialTestsFailure()
+                : new testActions.SendCompletedTestsFailure();
             });
           }),
         );
     }),
   );
 
+  @Effect()
+  sendPartialTestsSuccessEffect$ = this.actions$.pipe(
```
buzzub: The effect is plural but the action is singular so should this be `sendPartialTestSuccessEffect` instead?

buzzub:
Will approve once Robert's comments have been addressed 👍

src/providers/test-submission/test-submission.ts
```diff
@@ -37,10 +39,16 @@ export class TestSubmissionProvider {
...
+    const deepClonedData = cloneDeep(testToSubmit.payload);
+    const cleanData = this.removeNullFieldsDeep(
+      testToSubmit.status === TestStatus.WriteUp
+        ? this.removeFieldsForPartialData(deepClonedData)
+        : deepClonedData);
+
+    return this.httpClient.post(
+      this.buildUrl(testToSubmit.status),
+      this.compressData(cleanData),
       // Using cloneDeep() to prevent the initialState of the reducers from being modified
```
TomBillingtonUK: This comment is now in the wrong place

roberthall12 (approved):

buzzub (approved):

PR #774, opened 2019-09-19
MES-3612 Dispatch status update when leaving page

edwardwilson:
## Description

The status of the test is not moved to write up on the non-pass-finalisation page like it is on the pass page. This is now fixed and works like for like on both outcomes.


szabikr (commented):
Unit tests could be implemented for this change.

szabikr (approved):

rstepnowski (approved):

PR #783, opened 2019-09-23
MES-3535: Remote Config Schema Validation

szabikr:
## Description

Using `enjoi` and `joi` to validated the incoming remote configuration schema.

In order to be able to send logs when remote config schema is invalid we need to separate mdm config and remote config init. This PR contains that piece of work as well.

## Checklist

## Screenshots (optional)

IMG_0134

Log in case of schema validation error
Screenshot 2019-09-25 at 15 32 29


src/providers/app-config/app-config.ts
```diff
@@ -88,18 +91,34 @@ export class AppConfigProvider {
...
   public loadRemoteConfig = (): Promise<any> =>
     this.getRemoteData()
+      .then((data: any) => {
+        const result: ValidationResult<any> = this.schemaValidatorProvider.validateRemoteConfig(data);
+
+        if (result.error === null) {
+          return data;
+        }
+
+        return Promise.reject(result.error);
```
TomBillingtonUK: Personally, I would prefer the if statement to return the error and the return at the bottom to be the happy path. It being the other way round made it harder to read the first time

TomBillingtonUK (approved):

rossfauldsbjss (approved):
looks good

src/providers/app-config/app-config.ts
szabikr: Okay, good idea. Changing it

PR #798, opened 2019-09-26
Feature mes 3580 welsh competencies

rossfauldsbjss:
Use the real welsh translations for debrief competencies.

This does not cover all text that needs converting to welsh (this will be dealt with in another story)


amobarak (approved):

src/assets/i18n/cy.json
```diff
@@ -18,57 +18,57 @@
...
+      "judgementMeeting": "Doethineb - dod i gyfarfod cerbydau",
+      "judgementCrossing": "Doethineb - croesi",
+      "positioningNormalDriving": "Lleoli ar y ffordd - gyrru arferol",
+      "positioningLaneDiscipline": "Lleoli ar y ffordd - disgyblaeth lôn",
+      "clearance": "Clirio",
+      "followingDistance": "Pellter dilyn",
+      "useOfSpeed": "Defnyddio cyflymder",
+      "progressAppropriateSpeed": "Symud ymlaen - cyflymder addas",
+      "progressUndueHesitation": "Symud ymlaen - petruso gormodol",
+      "responseToSignsTrafficSigns": "Ymateb I arwyddion ffyrdd/ signalau - arwyddion traffig",
```
buzzub, line 52: There should probably be a space before the `/` in these translations e.g. `"Ymateb I arwyddion ffyrdd / signalau - arwyddion traffig"`

buzzub: Also, I would question why the Welsh translations don't use the same case. e.g.

English is "Response to signs / signals - Traffic signs"

Welsh is "Ymateb I arwyddion ffyrdd/ signalau - arwyddion traffig" but why not "Ymateb I arwyddion ffyrdd/ signalau - **A**rwyddion traffig" (capital "A")

rossfauldsbjss: Not sure the rules on  capitalisation are the same in welsh.  I did see an article that said they are to an extent - but less things are capitalised and thought it best not to second guess and just cut and paste exactly as provided.

TomBillingtonUK (approved):

rstepnowski (approved):

src/assets/i18n/cy.json
rossfauldsbjss: Added space before /  as looks better, and some of them had the space in already.

PR #803, opened 2019-09-30
MES-3636: Handling Different Categories in NGRX Store

szabikr:
## Description

Preparation for handling multiple categories. Implementing the approach in this PR.

The approach is that a reducer factory for each reducer will determine which category reducer to use depending on the started test category. The default reducer in all cases will be the already implemented cat B reducer so that we need to introducer a reducer only in the case of actual change.

This change includes:
- Refactor of the test data reducer, making it easier to work with and with more tests then ever before
- Ability to start a Cat B+E Test from the journal when running locally
- A `testDataReducerFactory` to allow us to use a different testDataReducer for each category
- A `instructorDetailsReducerFactory` which makes use of a new `nullReducer` which we can use where a category does not have that data

_To update the Category BE reducer we just create a new cat-be folder under the sub reducer we want to edit and create a copy of the default reducer and modify as needed. We should be able to reuse actions across reducers and we shouldn't need to have category-specific actions_

## Checklist


ammarhaiderbjss:
@szabolcs-keresztes needs to get merged into https://github.com/dvsa/mes-mobile-app/tree/mes-607-cat-be please

src/modules/tests/test-data/manoeuvres/manoeuvres.actions.ts
```diff
@@ -0,0 +1,62 @@
...
+export class AddManoeuvreDangerousFault implements Action {
+  constructor(public payload: ManoeuvrePayload) { }
+  readonly type = ADD_MANOEUVRE_DANGEROUS_FAULT;
+}
+
+export class AddManoeuvreComment implements Action {
+  constructor(
+    public fieldName: string,
+    public faultType: CompetencyOutcome,
+    public controlOrObservation: string,  // 'Control' | 'Observation',
```
rstepnowski, line 46: Would it be better to use a string literal type here instead of using comments explaining what we want to achieve here? e.g:
```controlOrObservationType = "Control" | "Observation";```

File: src/modules/tests/test-data/test-data.cat-be.reducer.ts [hunk 2]
```diff
@@ -0,0 +1,33 @@
+import { TestData } from '@dvsa/mes-test-schema/categories/B';
```
rstepnowski, line 1: I'm assuming that this will be replaced with B+E's TestData once schema is finalised, right?
Should this be the pattern moving forward for the rest of the categories?

src/modules/tests/test-data/controlled-stop/controlled-stop.reducer.ts
```diff
@@ -0,0 +1,47 @@
...
+      };
+    case controlledStopActions.CONTROLLED_STOP_ADD_DANGEROUS_FAULT:
+      return {
+        ...state,
+        fault: CompetencyOutcome.D,
+        selected: true,
+      };
+    case controlledStopActions.CONTROLLED_STOP_REMOVE_FAULT:
+      return {
+        selected: state.selected,
```
rstepnowski, line 37: Please remind since I'm not sure atm - you can have multiple types of faults on Controlled Stop? eg DF + D? Or just one type?
Will this properly remove 1 fault and not all?

rstepnowski (approved):
Amazing work!! Thanks @Cavalier07 & @szabolcs-keresztes !
Most of the comments are my questions/doubts or small improvements - good stuff :+1:

src/modules/tests/test-data/manoeuvres/manoeuvres.actions.ts
TomBillingtonUK: Yes, it would, during all of this we have left the initial implementation of the feature. I will happily make this change in a separate PR as this one is doing too much as it is

src/modules/tests/test-data/test-data.cat-be.reducer.ts
TomBillingtonUK: Yes , now the schema is published will make this change now :)

TomBillingtonUK: On second thoughts this is going to need the union pr merged so we only have to do this once , as we would need to refactor the app to use the extra schema. 

The aim would be this is replaced. :)

src/modules/tests/test-data/controlled-stop/controlled-stop.reducer.ts
TomBillingtonUK: With controlled stops, we allow 1 fault of any type (DF, Serious or Dangerous) and that is saved to a fault property. So when we remove the fault the only thing we need to keep is if it had been ticked.

Object is 

export interface ControlledStop {
  selected?: ManoeuvreIndicator;
  fault?: ManoeuvreOutcome;
  faultComments?: FaultComments;
}

MarkPatt89 (approved):
Waow... Awesome...

src/modules/tests/test-data/test-data.cat-be.reducer.ts
rstepnowski: I'm OK with leaving this as it is for now - just wanted to highlight it and clarify 👍

amobarak (approved):

PR #835, opened 2019-10-11
DRAFT - DO NOT MERGE - Dealing with multiple categories in state

TomBillingtonUK:
## Description
**This PR does not pass lint or tests - DO NOT MERGE**

- Refactor the tests.reducer to use a factory method
- Create a Cat B and B+E tests.reducer.ts

A number of other files have been changed so I can compile the code, each one should have a TODO

NOTES
- This method should allow us to only need one factory at the top level of the state as everything else can be hard-coded.
- There is a bug where the default factory (Cat B) will be called as we don't know what the category is in the test reducer at the beginning. We should pass it into the start test actions (all 11 of them to solve this) or refactor the start test effects  
- There is an opportunity to refactor test-data to work like journal data meaning we can  potentially have all the customization in one file for each category
- We should refactor the journal data object into a journal data folder like test data.

Thoughts and feedback welcome :)

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -554,9 +554,9 @@
       "dev": true
     },
     "@dvsa/mes-test-schema": {
-      "version": "3.0.0",
-      "resolved": "https://registry.npmjs.org/@dvsa/mes-test-schema/-/mes-test-schema-3.0.0.tgz",
-      "integrity": "sha512-MR4VJR7gsBB5f7rW8d99XAyT8+tlcVJSNDsPjJoeX61WFS1mWN40xyLSO+O8FrpJyXhRSlv+UGDQFjcH4a0T3w=="
+      "version": "3.1.0",
+      "resolved": "https://registry.npmjs.org/@dvsa/mes-test-schema/-/mes-test-schema-3.1.0.tgz",
+      "integrity": "sha512-5q0ROckyrvobWaXgVlxhcHiNWcW5zucIgJt2iUk6XZYzfN1uYZLlCny6rRnxNPOyIOsjnvzra/39c2FCV6IjHQ=="
     },
     "@fimbul/bifrost": {
       "version": "0.15.0",
@@ -1050,8 +1050,7 @@
             "ansi-regex": {
               "version": "2.1.1",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "aproba": {
               "version": "1.2.0",
@@ -1072,14 +1071,12 @@
             "balanced-match": {
               "version": "1.0.0",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "brace-expansion": {
               "version": "1.1.11",
               "bundled": true,
               "dev": true,
-              "optional": true,
               "requires": {
                 "balanced-match": "^1.0.0",
                 "concat-map": "0.0.1"
@@ -1094,20 +1091,17 @@
             "code-point-at": {
               "version": "1.1.0",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "concat-map": {
               "version": "0.0.1",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "console-control-strings": {
               "version": "1.1.0",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "core-util-is": {
               "version": "1.0.2",
@@ -1224,8 +1218,7 @@
             "inherits": {
               "version": "2.0.3",
               "bundled": true,
-              "dev": true,
-              "optional": true
+              "dev": true
             },
             "ini": {
               "version": "1.3.5",
@@ -1237,7 +1230,6 @@
               "version": "1.0.0",
               "bundled": true,
               "dev": true,
-              "optional": true,
               "requires": {
                 "number-is-nan": "^1.0.0"
               }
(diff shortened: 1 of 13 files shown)


package-lock.json
```diff
@@ -5316,9 +5297,9 @@
           "dev": true
         },
         "fsevents": {
-          "version": "2.0.7",
-          "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-2.0.7.tgz",
-          "integrity": "sha512-a7YT0SV3RB+DjYcppwVDLtn13UQnmg0SWZS7ezZD0UjnLwXmy8Zm21GMVGLaFGimIqcvyMQaOJBrop8MyOp1kQ==",
+          "version": "2.1.0",
```
ammarhaiderbjss, line 5300: Don't think we need this change?

src/modules/tests/tests-reducer-factory.ts
```diff
@@ -0,0 +1,20 @@
+import { TestCategory } from '../../shared/models/test-category';
+import { Action } from '@ngrx/store';
+import { StandardCarTestCATBSchema } from '@dvsa/mes-test-schema/categories/B';
+import { StandardTrailerTestCATBESchema } from '@dvsa/mes-test-schema/categories/BE';
+import { testsCatBReducer } from './tests.cat-b.reducer';
+import { testsCatBeReducer } from './tests.cat-be.reducer';
+
+type unionType = Required<StandardCarTestCATBSchema> | Required<StandardTrailerTestCATBESchema>;
+
+export function testsReducerFactory(category: string | null, action: Action, state: unionType): unionType {
```
ammarhaiderbjss, line 10: Nice!

src/modules/tests/tests.cat-b.reducer.ts
```diff
@@ -0,0 +1,71 @@
...
+  action: Action,
+  state: StandardCarTestCATBSchema,
+  ): Required<StandardCarTestCATBSchema> {
+  console.log('Cat B');
+  return {
+    ...nestedCombineReducers(
+      {
+        version: schemaVersionReducer,
+        category: categoryReducer,
+        activityCode: () => '1' as ActivityCode, // TODO - Need to fix
```
ammarhaiderbjss, line 40: Thought this wouldn't have needed to change? Can you explain?

package-lock.json
TomBillingtonUK: probably not - this pr won't go anywhere near one of the base branches , so I'm not too bothered about this one :)

src/modules/tests/tests.reducer.ts
```diff
@@ -89,55 +63,18 @@ const deriveSlotId = (state: TestsModel, action: Action): string | null => {
...
-          // the TS 'Required' mapped type which The 'Required' type which strips ? modifiers
-          // from all properties of 'StandardCarTestCATBSchema', thus making all properties required.
-          state.startedTests[slotId] as Required<StandardCarTestCATBSchema>,
-          action,
-        ),
+        // TODO - Need a way to get category - it needs to be in all the start test actions
+        // or we need to select it from the journal state
+        // ...testsReducerFactory('B', action, state.startedTests[slotId]),
+        // ...testsReducerFactory('B+E', action, state.startedTests[slotId]),
+        ...testsReducerFactory(get(state.startedTests[slotId], 'category'), action, state.startedTests[slotId]),
```
ammarhaiderbjss, line 77: Category should be simple to plumb through no?
And blimey, this is cool! Have we tested that this won't balloon our reducer calls when actions are dispatched?

ammarhaiderbjss:
Is this PR still a WIP? If so can we add it to the title?

src/modules/tests/tests.cat-b.reducer.ts
TomBillingtonUK: Currently, the ability to set the activity code is higher up in the reducer tree and works differently to anything else. Previously we only required the one object however now everything is typed Required<schema> it throws an error if the activity code is missing from the combine reducers. We just need to refactor the set activity code action so it's the same as the rest to fix this

src/modules/tests/tests.reducer.ts
TomBillingtonUK: It will be, just need to make a decision on how to plumb it through, I am not really keen and putting a selector into the journal module here. So I think we need to do one of the following:

1) Refactor all the start test actions to send a category
2) Move the current start test actions to hang of the set category action so start test -> set category -> all other actions

src/modules/tests/tests-reducer-factory.ts
```diff
@@ -0,0 +1,20 @@
+import { TestCategory } from '../../shared/models/test-category';
+import { Action } from '@ngrx/store';
+import { StandardCarTestCATBSchema } from '@dvsa/mes-test-schema/categories/B';
+import { StandardTrailerTestCATBESchema } from '@dvsa/mes-test-schema/categories/BE';
+import { testsCatBReducer } from './tests.cat-b.reducer';
+import { testsCatBeReducer } from './tests.cat-be.reducer';
+
+type unionType = Required<StandardCarTestCATBSchema> | Required<StandardTrailerTestCATBESchema>;
```
rossfauldsbjss, line 8: Really nice. However I think unionType should be something more meaningful like schemaUnionType

src/modules/tests/tests.reducer.ts
ammarhaiderbjss: First one makes sense, would the second one involve effects? I'd really like to avoid them!

TomBillingtonUK: The whole Start Test process is based on effects , so you can't avoid them

TomBillingtonUK: but yeah the second idea involves changing the effects structure, the first idea just changes the actions the effects create

ammarhaiderbjss:
@Cavalier07 can this be closed?

PR #836, opened 2019-10-11
MES-3783 use 10 minutes instead of slot duration to calculate when rekey can start

rossfauldsbjss:
allow rekeys to take place within 10 minutes of the test starting.


ammarhaiderbjss (approved):
Looks good, if we have time perhaps a quick unit test?

MarkPatt89 (approved):

rstepnowski (approved):

PR #852, opened 2019-10-16
MES-3798: Getting rid of nested combined reducers

szabikr:
## Description

Getting rid of nested combined reducers as we are not using it anymore

## Checklist


rossfauldsbjss (approved):
was nothing actually using it?
looks good.

MarkPatt89 (approved):

TomBillingtonUK (commented):
You need to update package lock file

TomBillingtonUK:
@Ross - we were using it but the latest store refactoring has allowed us to flatten the reducers so we aren’t nesting them any more. For example a Journal Data reducer has been added so we don’t nest those reducers in the tests.cat-b.reducer.ts

ross:
I think you want @rossfauldsbjss rather than @ross 😁

rstepnowski (approved):

PR #892, opened 2019-10-30
MES-3766 - Resume B+E test

TomBillingtonUK:
## Description
- Resuming a B+E Test now takes you to the Cat B+E versions of pages

## Checklist


rossfauldsbjss (approved):
looks good

MarkPatt89 (approved):

src/components/test-slot/test-outcome/test-outcome.ts
```diff
@@ -253,4 +253,22 @@ export class TestOutcomeComponent implements OnInit {
         return CAT_BE.WAITING_ROOM_PAGE;
     }
   }
+
+  getPassFinalisationPage(): string {
```
szabikr, line 257: Provider that gets the category from the store and returns the relevant page name would be a good shout here instead of the local functions. Especially `TestOutcome`, `TestOutcome` is already does way too much.

TomBillingtonUK: Happy to extract however I would suggest a helper class with these functions which take the category as a parameter.  As we don't connect providers to a store , an effect , new actions etc seem a bit overkill. What are your thoughts?

ammarhaiderbjss (approved):

PR #904, opened 2019-11-07
use staffNumber from test when authenticated staffNumber is null

rossfauldsbjss:
Populate staffNumber from the journal when authenticated staffNumber is null to fix issues of tests stuck at pending.

### DIFF
--- a/src/environment/test-schema-version.ts
+++ b/src/environment/test-schema-version.ts
@@ -1,1 +1,1 @@
-export const version = '3.2.4';
+export const version = '3.2.5';
--- a/src/modules/tests/tests.effects.ts
+++ b/src/modules/tests/tests.effects.ts
@@ -133,7 +133,7 @@ export class TestsEffects {
     switchMap(([action, journal, rekeySearch]: [testActions.StartTest, JournalModel, RekeySearchModel]) => {
       const startTestAction = action as testActions.StartTest;
       const isRekeySearch = this.navigationStateProvider.isRekeySearch();
-      const employeeId = this.authenticationProvider.getEmployeeId();
+      const employeeId = this.authenticationProvider.getEmployeeId() || journal.examiner.staffNumber;
 
       let slot: TestSlot;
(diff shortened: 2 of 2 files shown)


src/environment/test-schema-version.ts
```diff
@@ -1 +1 @@
-export const version = '3.2.4';
+export const version = '3.2.5';
```
ammarhaiderbjss, line 1: Was this necessary as part of this hotfix?

faggc (approved):

src/modules/tests/tests.effects.ts
```diff
@@ -133,7 +133,7 @@ export class TestsEffects {
     switchMap(([action, journal, rekeySearch]: [testActions.StartTest, JournalModel, RekeySearchModel]) => {
       const startTestAction = action as testActions.StartTest;
       const isRekeySearch = this.navigationStateProvider.isRekeySearch();
-      const employeeId = this.authenticationProvider.getEmployeeId();
+      const employeeId = this.authenticationProvider.getEmployeeId() || journal.examiner.staffNumber;
```
ammarhaiderbjss, line 136: With the absence of unit tests, assuming this has all been tested and verified manually?

src/environment/test-schema-version.ts
rossfauldsbjss: I didn't change this. somehow it changed itself. Believe Carl is doing something to remove this from the repository and git ignore it.

src/modules/tests/tests.effects.ts
rossfauldsbjss: Originally I created a helper method that wrapped this change - and had unit tests to prove that the correct value was returned if staffNumber was populated and if it was null. But realised that all I was really testing was that fundamental javascript behaviour worked as expected so there was no value to the test. 

Tested and verified manually.

ammarhaiderbjss (approved):

PR #935, opened 2019-11-20
[WIP] MES-3445 Bank holiday journal navigation

matthew2564:
## Description

Check for slot data past selectedDate in journal, and if any slot data exists allow navigation to the next date

## Checklist

### DIFF
--- a/src/modules/journal/__tests__/journal.selector.spec.ts
+++ b/src/modules/journal/__tests__/journal.selector.spec.ts
@@ -20,6 +20,7 @@ import { DateTimeProvider } from '../../../providers/date-time/date-time';
 import { DateTimeProviderMock } from '../../../providers/date-time/__mocks__/date-time.mock';
 import { Store } from '@ngrx/store';
 import { cloneDeep } from 'lodash';
+import { baseJournalData } from '../__mocks__/journal-slots-data.mock';
 
 class MockStore { }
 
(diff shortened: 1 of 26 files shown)


src/modules/journal/__tests__/journal.selector.spec.ts
```diff
@@ -56,6 +63,105 @@ describe('JournalSelector', () => {
     examiner: { staffNumber: '123', individualId: 456 },
   };
 
+  const baseJournalData: JournalModel = {
```
szabikr: Move this data into a mock file and import it, rather than defining it here.

src/modules/journal/__tests__/journal.selector.spec.ts
```diff
@@ -349,4 +455,16 @@ describe('JournalSelector', () => {
     });
   });
 
+  describe('hasSlotsAfterSelectedDate', () => {
+    it('should return TRUE if slots DO EXIST after journal selected date', () => {
+      expect(hasSlotsAfterSelectedDate(baseJournalData)).toEqual(true);
+    });
+
+    it('should return FALSE if slots DO EXIST after journal selected date', () => {
+      const journalWithoutSlotsAfterSelectedDate = cloneDeep(baseJournalData);
```
szabikr, line 366: use 
```
= {
...baseJournalData,
};
```
instead of `cloneDeep`

src/modules/journal/journal.selector.ts
```diff
@@ -28,6 +27,23 @@ export const canNavigateToPreviousDay = (journal: JournalModel, today: DateTime)
   return selectedDate > lookbackLimit;
 };
 
+/**
+ * Function to check if there is data after selected date in the journal slots, and to
+ * return a true signifying that the user can navigate back a day if that is the case.
+ * @param {JournalModel} journal
+ * @return {boolean}
+ */
+export const hasSlotsAfterSelectedDate = (journal: JournalModel): boolean => {
```
szabikr, line 30: As our convention goes: we try to write clean, understandable code and not comments around our functions. The code should be the documentation. You selected a very clear and understandable function name, input names, and we have the return type visible. That is enough for us to understand what this function does.

In the implementation, try to be more verbose and make sure that everything makes sense for the person who reads the code. 

If you do these things then you don't need to write comments, because it's just a maintenance overhead and weirdly elevate the complexity of the code.

ammarhaiderbjss:
Please add a description

src/modules/journal/__tests__/journal.selector.spec.ts
Cullima: There was a reason we chose cloneDeep over spread in this instance. Spread will only create a shallow copy and changes to nested objects in copies created using this method would therefore also undesirably update the original baseJournalData object.

PR #939, opened 2019-11-21
Feature mes 3949 implement catbe behaviour map

rossfauldsbjss:
Added a CatBE behaviour map, derived from the CatB but with vehicleChecks replacing showme and tellme mappings.

Created a selector to determine if vehicle checks card (to be created in a future PR) should be displayed or not based on behaviour map.

Changed cat be office page to use Cat BE map.


TomBillingtonUK (approved):

src/pages/debrief/cat-be/debrief.cat-be.selector.ts
```diff
@@ -131,6 +131,21 @@ export const getVehicleCheckDrivingFaults =
     return result;
   };
 
+export const vehicleChecksExist = (vehicleChecks: CatBEUniqueTypes.VehicleChecks): boolean => {
```
szabikr, line 136: Why do we need to check if vehicle checks exist? We can't progress the WRTC page if they don't and this is on debrief.

szabikr (commented):
looks good, but don't need that function in my opinion

szabikr:
If you are worried that it might be null and cause an uncaught exception use lodash get to get the show/tell me outcomes

src/pages/debrief/cat-be/debrief.cat-be.selector.ts
rossfauldsbjss: It is possible to end a test without entering anything against vehicle checks. This selector is used to check if we have a response or not and ultimately this will display the vehicle checks component or not.

ammarhaiderbjss, line 136: Needs a unit test

rossfauldsbjss: refactored to use lodash as agreed for clarity.

rossfauldsbjss: agreed. Will add unit tests.

src/providers/outcome-behaviour-map/outcome-behaviour-map.ts
```diff
@@ -38,6 +38,7 @@ export class OutcomeBehaviourMapProvider {
       return false;
     }
     if (field.display === VisibilityType.VisibleIfPresent && value) {
+      console.log(`value is ${value}`);
```
ammarhaiderbjss: console log

src/pages/office/office-behaviour-map.cat-be.ts
```diff
@@ -0,0 +1,648 @@
+import { OutcomeBehaviourMapping } from '../../providers/outcome-behaviour-map/outcome-behaviour-map.model';
+
+export const behaviourMap: OutcomeBehaviourMapping = {
```
ammarhaiderbjss, line 3: Out of interest, is this mapping documented anywhere? In confluence / Jira?

src/providers/outcome-behaviour-map/outcome-behaviour-map.ts
rossfauldsbjss: Will remove

src/pages/office/office-behaviour-map.cat-be.ts
rossfauldsbjss: There was a spreadsheet for cat b, think was against original ticket in jira. 
No Cat BE spreadsheet, but we have copied from Cat B and used same logic on VehicleChecks as used on tellMe question.

ammarhaiderbjss: Ok just as long as we're sure this displaying logic is uniform across categories

ammarhaiderbjss (approved):

appleabs (approved):

PR #964, opened 2019-11-26
bug fix - remove the traverse weekend logic from bank holiday navigation

matthew2564:
## Description
remove the traverse weekend code for journal navigation which was allowing the user to navigate to dates in future that had no slot data
## Checklist


src/modules/journal/journal.selector.ts
```diff
@@ -41,17 +41,10 @@ export const hasSlotsAfterSelectedDate = (journal: JournalModel): boolean => {
...
   const dayAfterTomorrowAsDate = DateTime.at(DateTime.today()).add(2, Duration.DAY).format('YYYY-MM-DD');
-  const nextDayAsDay = DateTime.at(journal.selectedDate).add(1, Duration.DAY).day();
 
-  // if the current day is a Friday(5) or Saturday(6), allow navigation to the Monday.
-  if ((DateTime.at(DateTime.today()).day() === 5 || DateTime.at(DateTime.today()).day() === 6) && nextDayAsDay !== 2) {
-    traverseWeekend = true;
-  }
-
-  return (nextDayAsDate < dayAfterTomorrowAsDate) || traverseWeekend || hasSlotsAfterSelectedDate(journal);
+  return (nextDayAsDate < dayAfterTomorrowAsDate) || hasSlotsAfterSelectedDate(journal);
```
szabikr, line 47: Should it only be the `hasSlotsAfterSelectedDate`? Can really think of a scenario when we would navigate to the next day and there are no slots on that day.

ammarhaiderbjss:
@matthew2564 Assuming this has been thoroughly tested against a range of journal scenarios?

src/modules/journal/journal.selector.ts
ammarhaiderbjss: I thought the user would need to be able to **navigate** to the next day even if there are no slots. Otherwise how would they verify that they have no slots?

szabikr: Uhh you are right. A weekend with no slots would be the best example.. Not sure how I missed that

matthew2564: So am i right in saying that the first check is correct then?

ammarhaiderbjss (approved):
Looks good provided we've tested against a range of journal scenarios

rossfauldsbjss (approved):
looks good

PR #970, opened 2019-11-26
Consolidates scss for all pages

ammarhaiderbjss:
## Description

- A diff was performed between all .scss files before creating the unified .scss files.
- Moves page .scss files up a level and removes page specific .scss files
- Uses class name as component selectors instead of element name
- Leverages the consistent class naming to apply scss based on an attribute wildcard selector 
- Had to tweak a couple of styling rules due to the change of using classes as component selectors now. Classes have a higher specificity weight than the previous element selectors. See https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity for more on this.

## Checklist


src/pages/test-report/components/timer/timer.scss
```diff
@@ -6,24 +6,27 @@ timer {
...
-    font-size: 17px;
-    font-weight: 500;
-    text-align: center;
-    letter-spacing: -0.35px;
+    span {
+      font-size: 17px;
+      font-weight: 500;
+      text-align: center;
+      letter-spacing: -0.35px;
+      color: #ffffff;
```
TomBillingtonUK, line 20: should be mes-white

src/pages/test-report/components/toolbar/toolbar.scss
```diff
@@ -54,17 +49,20 @@ toolbar {
     &.active {
       background: map-get($colors-gds, "gds-yellow");
     }
+    span {
+      font-size: 24px;
+      font-weight: 900;
+      text-align: center;
+      letter-spacing: -0.26px;
+      line-height: 25px;
+      color: #0b0c0c;
```
TomBillingtonUK, line 58: should be gds-black

TomBillingtonUK (approved):
Nice job - A smaller app is a happy app

rossfauldsbjss (approved):
Looks good - nice change. Does mean we will have to have Cat B and Cat BE specific stylings in the same file, but as these all get built in to one big css file I see no issue with that.

szabikr (approved):
Cool!

ammarhaiderbjss:
> Looks good - nice change. Does mean we will have to have Cat B and Cat BE specific stylings in the same file, but as these all get built in to one big css file I see no issue with that.

We could still add page specific .scss files in the future but I couldn't find anything that warranted one.

PR #979, opened 2019-11-29
MES-629 - Calculate Cat BE Test Result

TomBillingtonUK:
## Description
- Removed the temporary Pass and Fail options
- Added unit tests for existing functionality

## Checklist

### DIFF
--- a/src/providers/test-result/__tests__/test-result.spec.ts
+++ b/src/providers/test-result/__tests__/test-result.spec.ts
@@ -4,33 +4,12 @@ import { ActivityCodes } from '../../../shared/models/activity-codes';
 import { DrivingFaults } from '@dvsa/mes-test-schema/categories/Common';
 import { CatBUniqueTypes } from '@dvsa/mes-test-schema/categories/B';
 import { FaultCountProvider } from '../../fault-count/fault-count';
+import { TestCategory } from '../../../shared/models/test-category';
+import { CatBEUniqueTypes } from '@dvsa/mes-test-schema/categories/BE';
 
-// TODO: This test has tight coupling with category B
-// We will need to adjust it when we introduce new categories
 describe('TestResultCalculatorProvider', () => {
 
   let testResultProvider: TestResultProvider;
-
-  const initialTestData: CatBUniqueTypes.TestData = {
-    dangerousFaults: {},
-    drivingFaults: {},
-    manoeuvres: {},
-    seriousFaults: {},
-    testRequirements: {},
-    ETA: {},
-    eco: {},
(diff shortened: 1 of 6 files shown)


src/providers/test-result/__tests__/test-result.spec.ts
```diff
@@ -61,137 +40,216 @@ describe('TestResultCalculatorProvider', () => {
...
-      const testData: CatBUniqueTypes.TestData = {
-        ...initialTestData,
-      };
-      // ASSERT
-      testResultProvider.calculateCatBTestResult(testData).subscribe((result) => {
+      const testData: CatBEUniqueTypes.TestData = {};
+      testResultProvider.calculateTestResult(TestCategory.BE, testData).subscribe((result) => {
         expect(result).toBe(ActivityCodes.PASS);
         done();
       });
```
rossfauldsbjss, line 255: think there should be a test for anything over 15 driving faults with no serious and no dangerous.

src/providers/test-result/test-result.ts
```diff
@@ -11,13 +11,9 @@ import { CatBEUniqueTypes } from '@dvsa/mes-test-schema/categories/BE';
...
 
-  constructor(
-    private faultCountProvider: FaultCountProvider,
-  ) {}
-
-  calculateTestResult = (category: string,
-                         testData: CatBUniqueTypes.TestData| CatBEUniqueTypes.TestData): Observable<ActivityCode> => {
+  constructor(private faultCountProvider: FaultCountProvider) {}
 
+  public calculateTestResult(category: string, testData: object): Observable<ActivityCode> {
```
rossfauldsbjss, line 16: This does nothing if a category other than B or B+E is passed.  In some other cases i've seen it default to processing as cat b...  not sure if it should return a fail for anything it doesn't recognise??

rossfauldsbjss (approved):
looks good

src/providers/test-result/test-result.ts
ammarhaiderbjss: probably should `throw` something

ammarhaiderbjss (approved):
Nice

src/providers/test-result/__tests__/test-result.spec.ts
TomBillingtonUK: There is a test for that `should return a Fail when there are 16 driving faults'`

src/providers/test-result/test-result.ts
TomBillingtonUK: done

PR #993, opened 2019-12-04
fix issue with 16+ driving fault comments disappearing when one entered

rossfauldsbjss:
Test failures with no serious or dangerous faults  but 16+ driving faults display comment fields to the user on the office page to enter details into. 

Upon entering details in to any of these comment fields all of them disappeared.

Tracked this down to code that used Object.values combined with reduce.
Essentially it would add up the driving faults - however it picked up the comment field value, added that to the accumulated value (which resulted in a string instead of a number) and so failed the logic of  the result being > 15.

e.g . instead of a number it was returning "4<commentvalue>010"    

## Checklist

### DIFF
--- a/src/shared/helpers/__tests__/competency.spec.ts
+++ b/src/shared/helpers/__tests__/competency.spec.ts
@@ -1,0 +1,150 @@
+import { DrivingFaults, SeriousFaults, DangerousFaults } from '@dvsa/mes-test-schema/categories/Common';
+import { getCompetencyFaults, calculateFaultCount } from '../competency';
+// Driving Faults/ Serious Faults/ Dangerous faults
+
+describe('calculateFaultCount', () => {
+  it('should return count of 1 if boolean is passed', () => {
+    const result = calculateFaultCount(true);
+    expect(result).toBe(1);
+  });
+
+  it('should return value passed in if value is a mumber', () => {
+    const result = calculateFaultCount(4);
+    expect(result).toBe(4);
+  });
+  it('should return 0 if passed null', () => {
+    const result = calculateFaultCount(null);
+    expect(result).toBe(0);
+  });
+
+});
+describe('getCompetencyFaults', () => {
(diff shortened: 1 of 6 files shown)


src/shared/helpers/__tests__/competency.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+    const result = getCompetencyFaults(drivingFaults);
+
+    expect(result.length).toBe(3);
+    expect(result).toContain(
+      {
+        comment: null,
+        faultCount: 4,
+        competencyIdentifier: 'controlsAccelerator',
+        competencyDisplayName:'Controls - Accelerator',
+        source: 'simple',
```
TomBillingtonUK, line 61: Could you use the comment source enum here

src/shared/helpers/__tests__/competency.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+      };
+
+    const result = getCompetencyFaults(drivingFaults);
+
+    expect(result.length).toBe(3);
+    expect(result).toContain(
+      {
+        comment: null,
+        faultCount: 4,
+        competencyIdentifier: 'controlsAccelerator',
```
TomBillingtonUK: Could you use the Competencies enum here

TomBillingtonUK (approved):
Looks good, there are a couple of enums which could be used in the tests instead of strings.

src/shared/helpers/__tests__/competency.spec.ts
rossfauldsbjss: Good shout. Will do.

src/shared/helpers/__tests__/competency.spec.ts
rossfauldsbjss: Good shout. will do.

faggc (approved):

PR #1035, opened 2019-12-13
Hotfix/unit test fdescribe

ammarhaiderbjss:
## Description
- Uses lodash's `cloneDeep` when working with `initialState` in the vehicle checks reducer tests.
- The nested array structure of the state object and using the spread operator (`...initialState`) to copy state was causing cross contamination issues when running tests.

## Checklist

## Screenshots (optional)
image


TomBillingtonUK (approved):
Nice find :)

MarkPatt89 (approved):
Awesome work

src/pages/test-report/cat-be/components/vehicle-checks/vehicle-checks.scss
```diff
@@ -1,4 +1,4 @@
-.test-report-vehicle-checks {
+vehicle-checks {
```
szabikr, line 1: Should we name this component `vehicle-checks-office`? The selector is called that, but maybe we could name the files that way as well

edwardwilson (changes requested):
I would like to investigate the reducer further before we merge as I think there might be a run time bug here

edwardwilson (approved):
I have checked and the behaviour seen in the unit tests does not appear at runtime!

PR #1062, opened 2020-01-06
MES-4033 implement category C WRTC page

edwardwilson:
Implements WRTC page for the cat c schema as per the designs and required fields.
### Screenshot
image

### DIFF
--- a/src/pages/waiting-room-to-car/cat-c/waiting-room-to-car.cat-c.page.ts
+++ b/src/pages/waiting-room-to-car/cat-c/waiting-room-to-car.cat-c.page.ts
@@ -5,12 +5,8 @@ import { Store, select } from '@ngrx/store';
 import { StoreModel } from '../../../shared/models/store.model';
 import * as waitingRoomToCarActions from '../waiting-room-to-car.actions';
 import { Observable } from 'rxjs/Observable';
-import { GearboxCategory } from '@dvsa/mes-test-schema/categories/common';
 import { getCurrentTest, getJournalData } from '../../../modules/tests/tests.selector';
 import {
-  SchoolCarToggled,
-  DualControlsToggled,
-  GearboxCategoryChanged,
   VehicleRegistrationChanged,
 } from '../../../modules/tests/vehicle-details/common/vehicle-details.actions';
 import { map } from 'rxjs/operators';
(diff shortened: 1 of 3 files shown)


src/pages/waiting-room-to-car/cat-c/waiting-room-to-car.cat-c.page.ts
```diff
@@ -5,12 +5,8 @@ import { Store, select } from '@ngrx/store';
 import { StoreModel } from '../../../shared/models/store.model';
```
TomBillingtonUK, line 5: Are there no tests effected by the deleted functions, or were none of them being tested.

src/providers/fault-count/fault-count.ts
```diff
@@ -87,6 +88,36 @@ export class FaultCountProvider {
     };
   }
 
+  public getVehicleChecksFaultCountCatC = (vehicleChecks: CatCUniqueTypes.VehicleChecks): VehicleChecksScore => {
```
TomBillingtonUK, line 91: is there a way we can reuse the B+E function as this is an exact duplicate apart from the type. If not then can we have some unit tests for this function please

TomBillingtonUK (approved):
Looks good, just a couple of comments about unit testing

TomBillingtonUK:
Can we have a screenshot of the page as well, please with the removed fields

rossfauldsbjss (approved):
looks good baring in mind Tom's comments

src/pages/waiting-room-to-car/cat-c/waiting-room-to-car.cat-c.page.ts
edwardwilson: Nope, no unit tests affected

src/providers/fault-count/fault-count.ts
edwardwilson: I have not looked into this function right now as Vehicle checks come under another story (struggling to find right now, trying to find the number). I've just renamed the function so that the WRTC page references cat c specifics.

edwardwilson: MES-4037 is what covers the above. Refactoring will come under that ticket. Right now just needed to remove none CAT-C or common references

PR #1115, opened 2020-01-15
Feature mes 4421 cat a mod 1 speed check competency

MarkPatt89:
## Description
- Adding `speed-check-component` (containing `competency-button`, `input fields` and `action button`)
- Adding `speed-check-header-component` to allow the header to be displayed only once on the test report. (Header contains `First`, `Second` and `Speed requirement` labels)
- Note this ticket only adds UI properties. Implementation of logic will be added in `MES-4419`, I propose the unit tests will also be added in this as there is currently no functionality to test.
## Checklist

## Screenshots (optional)

Screenshot 2020-01-15 at 13 42 07

### DIFF
--- a/src/pages/test-report/cat-a-mod1/components/speed-check/__tests__/speed-check-header.page.spec.ts
+++ b/src/pages/test-report/cat-a-mod1/components/speed-check/__tests__/speed-check-header.page.spec.ts
@@ -1,0 +1,3 @@
+describe('SpeedCheckHeaderComponent', () => {
+  // todo: PREP-AMOD1 MES-4419 implement tests for component
+});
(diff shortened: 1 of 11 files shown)


ammarhaiderbjss:
Implementation looks really good dude! 👍 
I think I misunderstood you earlier, I thought you meant having two separate re-usable components. Looking at the code now I think you meant just having two usages of the same component. Is that right?
So I reckon it definitely should be just one component that gets re-used twice, once for emergency stop and once for avoidance..

Also, I'm sure you were already going to address this in your todo's but can we move this to the `/test-report/cat-a-mod1/components/` directory once its ready? Might need to create the directory..

faggc (approved):

rossfauldsbjss (approved):
looks good.

PR #1136, opened 2020-01-20
Mes 4056/debrief cat c

matthew2564:
## Description
Update debrief page for Cat C
## Checklist


src/pages/debrief/cat-be/components/vehicle-checks-card/__tests__/vehicle-checks-card.spec.ts
```diff
@@ -23,7 +23,7 @@ import { PopulateCandidateDetails }
 import { candidateMock } from '../../../../../../modules/tests/__mocks__/tests.mock';
 import { TestCategory } from '@dvsa/mes-test-schema/category-definitions/common/test-category';
 
-describe('VehicleChecksCardComponent', () => {
+describe('VehicleChecksCardComponentCatB', () => {
```
rossfauldsbjss: should be CatBE

src/providers/fault-summary/__tests__/fault-summary.spec.ts
```diff
@@ -164,6 +165,72 @@ describe('faultSummaryProvider', () => {
...
+              { outcome: 'DF' },
+              { outcome: 'DF' },
+            ],
+          },
+        };
+        const result = faultSummaryProvider.getDrivingFaultsList(data, TestCategory.C);
+        expect(result.length).toEqual(1);
+        expect(result[0].faultCount).toEqual(4);
+      });
+      it('should correctly return any vehicle checks faults when there are 5 driving faults', () => {
```
rossfauldsbjss: I'd like this test text to explicitly state that it would expect 4 faults

src/providers/fault-summary/fault-summary.ts
```diff
@@ -326,4 +391,60 @@ export class FaultSummaryProvider {
     }
     return result;
   }
+
+  private getVehicleCheckDrivingFaultsCatC(vehicleChecks: CatCUniqueTypes.VehicleChecks, category: CategoryCode)
```
rossfauldsbjss, line 395: think this will have an issue where there are 5 show me tell me question driving faults. 
This would result in it returning driving faults - rather than it seeing that there is a serious fault and exiting.

rossfauldsbjss: I think vehicleCheckFaults will have serious faults returned so this could be checked and returned.

rossfauldsbjss (changes requested):
looks good. think it needs additional check for serious fault in getVehicleCheckDrivingFaultsCatC  (for the case where vehicle checks has 5 driving faults  (should be considered a serious fault but isn't stored in serious faults)

src/assets/i18n/en.json
```diff
@@ -17,21 +17,71 @@
...
+        "S02": "Doors secure",
+        "S03": "Horn",
+        "S04": "Parking brake",
+        "S05": "Windscreen",
+        "S06": "Demist windscreen",
+        "S07": "Rear fog lights",
+        "S08": "Dipped to main beam",
+        "S09": "Brake lights"
+      },
+      "C": {
```
rstepnowski, line 37: I've asked Matt Murray to add the spreadsheet with the showMe/TellMe questions for all the vocational tests - if you want, you could include the all the questions in this PR or open another one (since there is a ticket for C1, C1+E, C+E subcategories)

src/assets/i18n/en.json
```diff
@@ -17,21 +17,71 @@
...
+        "T1": "Safety factors while loading",
```
rstepnowski, line 38: The removal of leading zeros from the question indicator - can you remind me what was the reason for it? Is it intentional since those TXX/SXX will be sent to RSIS so I'm not sure if that's expected on their end

edwardwilson: Also, I don't think it will be picked up by the component as it uses the code as the key when selecting the text in the JSON. The code has the leading zero...

src/providers/fault-summary/fault-summary.ts
matthew2564: @rossfauldsbjss i've replicated the logic that exists in BE, and the Cat BE/C have the same number of vehicle checks, so is the logic not the same?

src/assets/i18n/en.json
matthew2564: @rstepnowski so i got the codes from the constants file that exist already in the app (src/shared/constants/show-me-questions/show-me-questions.vocational.constants.ts), i assumed these should've aligned to the values in the translation JSON objects, and therefore copied the codes, if thats not the case should be minor changes

src/providers/fault-summary/__tests__/fault-summary.spec.ts
matthew2564: is that not this expectation `expect(result[0].faultCount).toEqual(4);` ?

src/providers/fault-summary/fault-summary.ts
rossfauldsbjss: It looks like a logic flaw to me - in that it exits if there is a serious or dangerous fault, but in the case of 5 df this is considered a serious.  I think the flaw is in both, however depending on what uses this method it may not manifest as an issue.  So not sure.  I think it looks to me like it would be used to bypass editing of fault comments on the office page.  Could you check what uses it please?

src/providers/fault-summary/__tests__/fault-summary.spec.ts
rossfauldsbjss: It's more that if someone wasn't familiar with the rule  they might change this to 5. it's just bringing the description into line with the test. So for example "Should correctly return 4 vehicle check faults when there are 5" ....

src/assets/i18n/en.json
edwardwilson: Think it's worth checking because I believe with B+E they were needed for the translation mapping etc

src/providers/fault-summary/__tests__/fault-summary.spec.ts
matthew2564: ah right the description, yes valid point will update that now

rossfauldsbjss (approved):
looks good

src/assets/i18n/en.json
rstepnowski: Yeah, we've asked that question to RSIS guys - they don't mind atm but for consistency sake, we want to remove the leading zeros to align all of the categories with cat B.

MarkPatt89 (approved):

PR #1138, opened 2020-01-20
MES 4699: Use enums for Pass/Fail

faggc:
## Checklist


ammarhaiderbjss (approved):

src/pages/debrief/cat-a-mod1/debrief.cat-a-mod1.page.ts
```diff
@@ -176,7 +178,7 @@ export class DebriefCatAMod1Page extends BasePageComponent {
   endDebrief(): void {
     this.store$.dispatch(new EndDebrief());
     // TODO: refactor to use an enum for outcome
```
szabikr, line 180: TODO can be removed

szabikr (approved):

rossfauldsbjss (approved):
looks good - However, I see nowhere using TestOutcome.FAIL?
Is there nowhere that sets th test outcome to a fail?

faggc:
@rossfauldsbjss not that I could see, at least from the particular strings I tested.
There are lower case pass/fail strings, but I believe that is for another purpose and we should review those separately.

PR #1143, opened 2020-01-21
[JAB]  fixed cat b failed journey test and added new test

jono-basford-bjss:
- Fixed the cat b failed user journey test

- Added a new test which asserts  when 5 driver faults are added during the show me/tell questions a serious fault is added and is visible at the test report page

-Added scenario outlines to feature files due to using data table to answer show me/tell questions


andrew-lai-bjss (changes requested):
I think we need to review the purpose of the step definition as it is now used in two places for different actions.

We probably need a new step definition for cat B+E and others and leave the original step def as cat B only as it follows a different process.

PR #1181, opened 2020-01-29
MES-4696 Feature/store refactor

edwardwilson:
## Description
Added barrel files into some of the store sections to improve imports and namespace categories better.
## Checklist


amobarak:
Don't see what this PR is trying to solve, I rather prefer to old version of the imports, they're more defined and easier to understand.

edwardwilson:
The existing way requires imports from many individual files, which allows for mistakes to happen by having multiple categories on one page. With the following changes, the import is more descriptive that of the 'area' we are bring type in from.

amobarak:
I honestly think this should be a cross team decision whether to change the imports or leave them the same as they are, as I still prefer the old way of importing.

src/modules/tests/journal-data/common/journal-data.reducar.ts
```diff
@@ -1,10 +1,11 @@
 import { JournalData } from '@dvsa/mes-test-schema/categories/common';
```
rossfauldsbjss: File needs a rename as misspelt reducer (reducar)

rossfauldsbjss (approved):
Not a huge fan of this change. Think it's a lot of changes for little benefit.
Also feel like this should go through some  thorough testing to ensure that 
released categories still work.

edwardwilson:
> Not a huge fan of this change. Think it's a lot of changes for little benefit.
> Also feel like this should go through some thorough testing to ensure that
> released categories still work.

Could you give some feedback on what specifically you aren't keen on?

I agree regression testing is always important but there is no function change here; it is all simply file references so the change is minimal.

rossfauldsbjss:
Okay - I don't see what issue it is actually addressing.  Does it make the code easier to read. No not particularly.  Does it address a major problem that is causing issues. Not as far as I can tell.  It touches a large number of files throughout the app - is the possibility of this change introducing side   effects outweighed by the benefits of implementing it. Again not as far as i can tell.  A sweeping change like this should at the very least have been discussed cross team - so that perhaps there is a better understanding of the proposed benefits of this change, and also to take part in the decision making.

TomBillingtonUK:
A couple of thoughts

The new import for getting Eyesight Test would be 
`import { EyesightTestFailed } from '../../../../modules/tests/test-data';`
Eyesight Test is one of those things which is common between some categories but not in all. We lose the clarity on if we should have an eyesight test or not per category. Do we need to rethink what we class as common, which to me is used in Two or More Categories?

This PR only fixes half the issue your trying to solve, you've changed the state side of the app but all the pages are still looking at specific files like the one below

`import { getEco } from '../../../../modules/tests/test-data/common/test-data.selector';`

Is there a plan to change all of that as if there are now 2 ways of importing the same thing. It adds to the confusion already in the app. The next category is a copy and paste which isn't going to make things any better. If we push this in, is there a way to make sure we can't do it the other way and force everyone to do it the new way (not export the files outside the module)

edwardwilson:
> A couple of thoughts
> 
> The new import for getting Eyesight Test would be
> `import { EyesightTestFailed } from '../../../../modules/tests/test-data';`
> Eyesight Test is one of those things which is common between some categories but not in all. We lose the clarity on if we should have an eyesight test or not per category. Do we need to rethink what we class as common, which to me is used in Two or More Categories?
> 
> This PR only fixes half the issue your trying to solve, you've changed the state side of the app but all the pages are still looking at specific files like the one below
> 
> `import { getEco } from '../../../../modules/tests/test-data/common/test-data.selector';`
> 
> Is there a plan to change all of that as if there are now 2 ways of importing the same thing. It adds to the confusion already in the app. The next category is a copy and paste which isn't going to make things any better. If we push this in, is there a way to make sure we can't do it the other way and force everyone to do it the new way (not export the files outside the module)

Completely right this only affects the state of the app which is why although looking like a large PR has little 'real' change, mainly just introducing lots of index.ts files.

The issue that is trying to be solved here is having a simple/singular place to import types from a specific place of the state. Therefore only one place to correctly or incorrectly get a category required over many individual imports allowing a mismatch of categories. Which has been found throughout the app due to the copy and paste nature of new categories.

In terms of going forward, there is no real way to 'force' a way of importing in the current app. That would require taking advantage of npm modules which is a different can of worms. Also changing the existing pages has not been done to minimise the sweeping change

szabikr:
I do like changes like this as it improves a core problem in the app which is the number of imports that each file has. They are difficult to read, there is no particular order to them and the fact that a character can alter the category is actually very scary (the team definitely likes to live dangerously).

On the other hand, yes a change like this should have been communicated to the team upfront and not just in a form of a PR. As we've seen a few comments here that could be very useful to bear in mind while making these changes.

In my opinion keeping the consistency is one of the most important practice that we can do in a large application like DES. So we have to make sure that we can define our import strategy and apply that in every single file.

szabikr:
This PR needs closing as the ticket containing the work didn't get into this sprint and probably won't get into next one either

PR #1185, opened 2020-01-29
Bug fix manoeuvres for reverse left

mgoldson:
## Description
The reverse Left component for CAT C was referencing CAT BE methods for selectors/reducers etc
Also, functionality to call the correct category specific methods was absent.

- Adds basic provider to pull test data via category code (Can be used for CAT D and sub cats also)
- Passes test category into the reverse left component
- Type unions used on test data selector to save duplicating whole selector files
- Reverse left component uses the test category to call correct state management methods

## Checklist

## Screenshots (optional)

Simulator Screen Shot - iPad Pro (10 5-inch) - 2020-01-29 at 17 55 44
Simulator Screen Shot - iPad Pro (10 5-inch) - 2020-01-29 at 17 55 56

### DIFF
--- a/src/pages/test-report/cat-c/components/reverse-left/__tests__/reverse-left.spec.ts
+++ b/src/pages/test-report/cat-c/components/reverse-left/__tests__/reverse-left.spec.ts
@@ -16,7 +16,11 @@ import { testsReducer } from '../../../../../../modules/tests/tests.reducer';
 import { testReportReducer } from '../../../../test-report.reducer';
 import { StartTest } from '../../../../../../modules/tests/tests.actions';
 import { ReverseLeftPopoverClosed, ReverseLeftPopoverOpened } from '../reverse-left.actions';
-import { FaultCountProvider } from '../../../../../../providers/fault-count/fault-count';
+import { DeselectReverseLeftManoeuvreCatC }
+  from '../../../../../../modules/tests/test-data/cat-c/manoeuvres/manoeuvres.cat-c.actions';
+import { RecordManoeuvresSelection }
+  from '../../../../../../modules/tests/test-data/common/manoeuvres/manoeuvres.actions';
+import { ManoeuvreTypes } from '../../../../../../modules/tests/test-data/test-data.constants';
 import { AppModule } from '../../../../../../app/app.module';
 import { TestCategory } from '@dvsa/mes-test-schema/category-definitions/common/test-category';
 import { configureTestSuite } from 'ng-bullet';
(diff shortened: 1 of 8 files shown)


src/pages/test-report/cat-c/components/reverse-left/__tests__/reverse-left.spec.ts
```diff
@@ -52,14 +53,43 @@ describe('reverseLeftComponent', () => {
     component = fixture.componentInstance;
     store$ = TestBed.get(Store);
     // TODO: MES-4287 use category C
```
rossfauldsbjss: remove TODOS if they no longer apply

edwardwilson:
Should the red and yellow fault indicators align to the right edge?

edwardwilson (changes requested):
Just the TODO comment Ross picked up, otherwise looks good 👍

rossfauldsbjss (approved):
looks good

mgoldson:
@edwardwilson TODO has been removed. Good question about the fault icon placement. You can't have both a serious and dangerous fault at the same times so feels like they should be right aligned. Will check on this with Design.

src/providers/test-data-by-category/test-data-by-category.ts
```diff
@@ -0,0 +1,20 @@
...
+
+@Injectable()
+export class TestDataByCategoryProvider {
+  public getTestDataByCategoryCode(category: CategoryCode) {
+    switch (category) {
+      case TestCategory.C: return getTestDataC;
+      case TestCategory.C1: return getTestDataC1;
+      case TestCategory.CE: return getTestDataCE;
+      case TestCategory.C1E: return getTestDataC1E;
+      default: return getTestDataC;
```
rstepnowski: Does this default is a safe thing to do here? 
Are we planning to reuse this provider in other categories (eg. D?) in the future? If yes, defaulting to C might be a risky thing ... even for one of the subcategories of C it seems a bit off imho...

rstepnowski (commented):
Everything seems fine but I'm not quite sure about the future of that provider that you created and whether or not is should throw an error when wrong category indicator is provided...

src/providers/test-data-by-category/test-data-by-category.ts
mgoldson: Yes, this will be extended for CAT D too. I'm happy to take the default out.

mgoldson:
@rstepnowski I have address your comment now.

rstepnowski (approved):

edwardwilson (approved):

PR #1199, opened 2020-01-31
Feature/mes 4669 mod2 category selection

edwardwilson:
## Description
Added the change category selector into the Mod2 WRTC page following the implementation for Mod1.

## Checklist

## Screenshots (optional)
image

### DIFF
--- a/src/pages/waiting-room-to-car/cat-a-mod1/waiting-room-to-car.cat-a-mod1.page.ts
+++ b/src/pages/waiting-room-to-car/cat-a-mod1/waiting-room-to-car.cat-a-mod1.page.ts
@@ -142,8 +142,8 @@ export class WaitingRoomToCarCatAMod1Page extends BasePageComponent {
         map(isManual),
       ),
       testCategory$: currentTest$.pipe(
-      select(getTestCategory),
-    ),
+        select(getTestCategory),
+      ),
     };
   }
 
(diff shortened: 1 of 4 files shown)


MarkPatt89 (approved):

Cullima (changes requested):
Think you need to update tests.effects to stop the populate test action from being fired on starting the test from the journal. This is how it was done for mod 1.

edwardwilson:
That is changing in a future PR as the change in tests.effects breaks the analytics. The fix for mod1 fixes it for all of them

Cullima:
> That is changing in a future PR as the change in tests.effects breaks the analytics. The fix for mod1 fixes it for all of them



> That is changing in a future PR as the change in tests.effects breaks the analytics. The fix for mod1 fixes it for all of them

Ah, ok. Does the page still present the button only prior to the category being confirmed? This was the reason for not dispatching the action initially. I think the idea is the user is forced to open the modal and confirm the category.

edwardwilson:
> > That is changing in a future PR as the change in tests.effects breaks the analytics. The fix for mod1 fixes it for all of them
> 
> > That is changing in a future PR as the change in tests.effects breaks the analytics. The fix for mod1 fixes it for all of them
> 
> Ah, ok. Does the page still present the button only prior to the category being confirmed? This was the reason for not dispatching the action initially. I think the idea is the user is forced to open the modal and confirm the category.

Right now the page loads with an existing selection. So it doesn't show the Button to select, only the option to change. In commit https://github.com/dvsa/mes-mobile-app/commit/b1a334cfd74d8cf34c49ccf63f9403d604c9c11c the changes to the bike category component can be seen which introduces the new functionality.

I don't think we should hold this PR up while we wait for the other changes to get merged. We can review the story after to confirm the AC has been met later

Cullima (approved):

PR #1229, opened 2020-02-05
[JAB] Automatic transmission check

jono-basford-bjss:
Adding an assertion which checks the Automatic Transmission displays in Search


Cullima (approved):

rossfauldsbjss (approved):
looks good

PR #1235, opened 2020-02-05
[JAB] Updated Tags

jono-basford-bjss:
Reduced the scope of the smoke tests and added a new tag for a scheduled build which runs more tests. 

Removed test:e2e-simulator-bdd-smoke from package.json as its not required


rossfauldsbjss (approved):
looks good

faggc (approved):

PR #1238, opened 2020-02-06
Bugfix/mes 4765 fixing vehicle checks trailer calculation

amobarak:
## Description

Vehicle checks scoring fixed for C+E, C1+E, D+E, D1+E, updated tests too.

## Checklist

### DIFF
--- a/src/providers/fault-count/cat-c/fault-count.cat-c.ts
+++ b/src/providers/fault-count/cat-c/fault-count.cat-c.ts
@@ -27,30 +27,6 @@ export class FaultCountCHelper {
     return FaultCountCHelper.getDangerousFaultSumCountNonTrailer(data);
   }
 
-  public static getVehicleChecksFaultCountCatC = (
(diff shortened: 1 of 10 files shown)


src/providers/fault-count/cat-c/fault-count.cat-c.ts
```diff
@@ -268,4 +244,28 @@ export class FaultCountCHelper {
     return result;
   }
 
+  public static getVehicleChecksFaultCountCatC = (
```
TomBillingtonUK, line 247: we could do with unit tests for all of these. As we don't seem to have any, all bug fixes should have unit tests to make sure that specific thing can't happen again

TomBillingtonUK: I noticed there isn't a spec file related to this one (there should be instead of us relying on the entry point provider to test everything)

appleabs (approved):

PR #1243, opened 2020-02-06
fix duplicate selector and duplicate slot ids in local mock

rossfauldsbjss:
No MES number for this. Just a quick fix of a couple of issues noticed after multiple PRs have been merged in.

1.  WRTC - vehicle checks button styling affected by test report changes. Fix was to de-duplicate the selector on the Test Report page.
2. When serving locally, we had multiple test slots with the same slotID resulting in their status all being the same (e.g Resume etc).

## Checklist

- [X] Branch is rebased against the latest develop
- [X] Code has been tested manually


src/pages/test-report/cat-d/components/vehicle-checks/vehicle-checks.cat-d.ts
```diff
@@ -17,7 +17,7 @@ interface ComponentState {
 }
 
 @Component({
-  selector: 'vehicle-checks-cat-d',
+  selector: 'vehicle-checks-test-report-cat-d',
```
szabikr, line 20: Can we change the file and the class name as well?

rossfauldsbjss: Good point. I'll change that now.

Cullima (approved):

faggc (approved):

TomBillingtonUK (approved):

PR #1271, opened 2020-02-11
Consolidating a number of UI tests into the full journeys to reduce execution times

andrew-lai-bjss:
Consolidated a number of tests by moving assertions into the full journey scenarios, reducing the number of tests and therefore the execution time but retaining the test assertions.

Also fixed validation UI checks issue due to page updates.


rossfauldsbjss (approved):
looks good

mgoldson (approved):

turonekt (approved):

PR #1277, opened 2020-02-12
remove NA option

rossfauldsbjss:
Show me questions for Cat C and sub categories appear with N/A as an option.
This can be chosen and a driving fault allocated it against it.

The fix is to remove N/A as an option for Cat C and subcats

## Checklist


TomBillingtonUK (approved):

amobarak:
Would be worth getting rid of it in the `cy.json` and `en.json` files too

Cullima (approved):

PR #1287, opened 2020-02-12
Adding UI test run data to HTML report

andrew-lai-bjss:
Updated UI tests to output some of the run config into the HTML report so it is clear what was run.

Note that the version is taken from the config.xml file so might not be accurate if it has not been updated.

### DIFF
--- a/generateTestReport.js
+++ b/generateTestReport.js
@@ -1,11 +1,38 @@
+var fs = require('fs');
+var xml2js = require('xml2js');
 var reporter = require('cucumber-html-reporter');
+
+var parser = new xml2js.Parser();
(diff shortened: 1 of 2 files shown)


andrew-lai-bjss:
Example where a feature file is specified
Screenshot 2020-02-12 at 17 02 10
Example where a tag was specified
Screenshot 2020-02-12 at 17 03 13
Example where a tag was specified and was executed against iOS 11.4
Screenshot 2020-02-12 at 17 05 09

jono-basford-bjss (approved):
🍍

turonekt (approved):

PR #1292, opened 2020-02-13
MES-4670, MES-4509 - feat(SafetyAndBalanceChecks): fully implement Safety and Balance chec…

AndyPerryKainos:
Fully implement Safety and Balance checks for Cat A Mod 2

## Checklist

## Screenshots (optional)
IMG_0005
IMG_0006
IMG_0007
IMG_0008
IMG_0009


faggc (approved):

src/shared/constants/balance-questions.cat-a-mod2.constants.ts
```diff
@@ -1,23 +1,22 @@
 import { VehicleChecksQuestion } from '../../providers/question/vehicle-checks-question.model';
 
-// TODO - PREP-AMOD2: Add balance questions (MES-4670)
 export default [
   {
-    code: 'T01',
+    code: 'Q01',
```
TomBillingtonUK: There have been discussions on other categories where we shouldn't have leading zeros in the questions so it should be Q1, Q2 ... Can you double check if we should or shouldn't here

TomBillingtonUK (approved):
Looks good, just one query on leading zeros

src/shared/constants/balance-questions.cat-a-mod2.constants.ts
AndyPerryKainos: Who is the best person to make this decision?

AndyPerryKainos:
Clarification from Matt Murray:
- Leading zeros removed
- Balance question code prefix is "BQ"
- Safety question code prefix is "SQ"

PR #1296, opened 2020-02-13
Tech/journal performance

matthew2564:
## Description
After some digging around, noticed that the canStartTest function in the Slot provider was causing the performance issue. Then subsequently traced the source of the problem to the use of moment in the DateTime class. The use of Moment was in a loop that was firing extensively and causing memory issues. After using the JS date functions instead started to notice vast improvements. Not sure if its worth investigating using a different version of Moment as a test to see if that has any bearing.
## Checklist

### DIFF
--- a/src/providers/slot/slot.ts
+++ b/src/providers/slot/slot.ts
@@ -73,7 +73,7 @@ export class SlotProvider {
       numberOfDaysToView,
       (d: number): string => this.dateTimeProvider.now().add(d, Duration.DAY).format('YYYY-MM-DD'),
     );
-    const emptyDays = days.reduce((days: { [k: string]: SlotItem[] }, day: string) => ({...days, [day]: []}), {});
+    const emptyDays = days.reduce((days: { [k: string]: SlotItem[] }, day: string) => ({ ...days, [day]: [] }), {});
 
     return {
       ...emptyDays,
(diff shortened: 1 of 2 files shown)


ammarhaiderbjss (changes requested):
@matthew2564 Could we please get a description of the changes and how it relate to tackling performance issues?

src/providers/slot/slot.ts
```diff
@@ -99,30 +102,34 @@ export class SlotProvider {
...
+      const slotHasPeriodEndCriteria: boolean = this.hasPeriodEndCriteria(slotStartDate, period.to);
       return period.testCategory === testCategory && slotHasPeriodStartCriteria && slotHasPeriodEndCriteria;
     });
     return periodsPermittingStart.length > 0;
   }
 
-  private hasPeriodStartCriteria(slotDate: DateTime, periodFrom: string) {
-    const periodStartDate = new DateTime(periodFrom);
-    return slotDate.daysDiff(periodStartDate) <= 0;
+  private dateDiffInDays = (startDate: Date, periodDate: Date): number => {
```
ammarhaiderbjss: Quick unit test to cover this new method?

ammarhaiderbjss (changes requested):
Just a unit test for `dateDiffInDays()` :)

src/providers/slot/slot.ts
matthew2564: shall i make it public in order to test it?

rossfauldsbjss (approved):
looks good

ammarhaiderbjss (approved):

PR #1317, opened 2020-02-20
MES-4922 Fixed showing aborted attempt in speed requirement

edwardwilson:
## Description
Fixed the code checks that broke when the data changed from string to numbers

## Checklist

### DIFF
--- a/src/modules/tests/test-data/cat-a-mod1/avoidance/avoidance.selector.ts
+++ b/src/modules/tests/test-data/cat-a-mod1/avoidance/avoidance.selector.ts
@@ -3,6 +3,6 @@ import { TestData, Avoidance } from '@dvsa/mes-test-schema/categories/AM1';
 export const getAvoidance = (testData: TestData): Avoidance => testData.avoidance;
 
 export const getAvoidanceAttempted = (avoidance: Avoidance) : boolean => {
-  if (avoidance.firstAttempt || avoidance.secondAttempt) return true;
+  if (avoidance.firstAttempt !== undefined || avoidance.secondAttempt !== undefined) return true;
   return false;
 };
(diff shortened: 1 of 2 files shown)


src/modules/tests/test-data/cat-a-mod1/avoidance/avoidance.selector.ts
```diff
@@ -3,6 +3,6 @@ import { TestData, Avoidance } from '@dvsa/mes-test-schema/categories/AM1';
 export const getAvoidance = (testData: TestData): Avoidance => testData.avoidance;
 
 export const getAvoidanceAttempted = (avoidance: Avoidance) : boolean => {
-  if (avoidance.firstAttempt || avoidance.secondAttempt) return true;
+  if (avoidance.firstAttempt !== undefined || avoidance.secondAttempt !== undefined) return true;
```
szabikr: you can return the expression evaluation and remove the if statement

src/pages/debrief/cat-a-mod1/components/speed-check-debrief-card/speed-check-debrief-card.html
```diff
@@ -1,4 +1,4 @@
-<ion-card id="speed-check" *ngIf="emergencyStop.firstAttempt || avoidance.firstAttempt">
+<ion-card id="speed-check" *ngIf="emergencyStop.firstAttempt !== undefined || avoidance.firstAttempt !== undefined">
```
szabikr, line 1: use the same function that is implemented in the selector

szabikr (approved):

rossfauldsbjss (approved):
Agree with Szabi's comments. 
Looks good.

src/pages/debrief/cat-a-mod1/components/speed-check-debrief-card/speed-check-debrief-card.html
edwardwilson: The selector checks if the avoidance was attempted, it doesn't check if the emergency stop was like above. There could be a function created if desired but the suggestion to use the selector wouldn't work.

src/modules/tests/test-data/cat-a-mod1/avoidance/avoidance.selector.ts
edwardwilson: Yes correct! We can now before there was a typical issue with them typed as strings but compared to undefiend we can use a lamba statment

PR #1322, opened 2020-02-21
MES-4118 Vehicle Check Faults Display Fix + Termination Error

tomcrawleyy:
## Description
**Issue**
If a DE selects a ShowMeTellMe Question but does not provide an outcome for the question before terminating, the question will show up as a fault on debrief page.
**Fix**
NgOnInit added to VehicleChecksCard to capture and remove any ShowMeTellMe provided without an outcome from the state.
**Explanation**
This in turn will mean that the vehicle checks card will not show if a DE selected a question and did not provide an outcome for that question before terminating the test. Previously; it would include the question as a fault if the outcome didn't match to a pass, therefore if the property did not exist it would mark it as a fault, meaning the vehicle checks card would show in instances where unnecessary/unwanted.
**Note**
If a DE selects 2 or more ShowMeTellMe Questions, and provides an outcome for only 1 of the questions, only the question with an outcome provided will display in the debrief page.

**Also fixed** 
Terminating the test on the WRTC page for Cat D tests will now point to the correct Debrief page for Cat D. 
## Checklist

### DIFF
--- a/src/pages/debrief/components/vehicle-checks-card/__mocks__/vehicle-checks-card.mock.ts
+++ b/src/pages/debrief/components/vehicle-checks-card/__mocks__/vehicle-checks-card.mock.ts
@@ -1,42 +1,28 @@
 import { QuestionResult } from '@dvsa/mes-test-schema/categories/common';
 
-export class VehicleChecksCardMock {
-
-  private static instance: VehicleChecksCardMock;
-
-  private constructor() { }
-
-  public static getInstance(): VehicleChecksCardMock {
-    if (!VehicleChecksCardMock.instance) {
-      VehicleChecksCardMock.instance = new VehicleChecksCardMock();
-    }
-    return VehicleChecksCardMock.instance;
-  }
-
-  public getMalformedVehicleChecks(): QuestionResult[] {
-    return [
-      {
-        code: 'T1',
-        description: 'Safety factors while loading',
-      },
-      {
-        code: 'T2',
-        description: 'Reflectors condition',
-        outcome: 'DF',
-      },
-      {
-        code: 'S1',
-        description: 'All doors secure',
-      },
-      {
-        code: 'S2',
-        description: 'Air leaks',
-        outcome: 'P',
-      },
-      {
-        code: 'S4',
-        description: 'Mudguards condition',
-      },
-    ];
-  }
-}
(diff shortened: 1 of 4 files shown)


TomBillingtonUK (commented):
Needs unit tests.

I'm assuming PR's will be following to add this to Cat C for develop, release-3.1.x and release-3.2.0

src/pages/debrief/components/vehicle-checks-card/vehicle-checks-card.ts
```diff
@@ -17,4 +17,12 @@ export class VehicleChecksCardComponent {
   constructor() { }
 
   questionHasFault = (result: QuestionResult): boolean => result.outcome !== CompetencyOutcome.P;
+
+  ngOnInit(): void {
+    this.tellMeShowMeQuestions.forEach((value: QuestionResult, index: number, array: QuestionResult[]): void => {
+      if (!value.hasOwnProperty('outcome')) {
+        this.tellMeShowMeQuestions.splice(index, 1);
+      }
```
TomBillingtonUK: Splicing an array as your iterating through it feels a bit risky to me. A .filter would be a lot safer to remove the elements instead of using a .foreach

tomcrawleyy: You are completely right, it'll skip an index after the splice.. I don't know how I missed that, nice spot, will be amended soon.

src/pages/debrief/components/vehicle-checks-card/__mocks__/vehicle-checks-card.mock.ts
```diff
@@ -0,0 +1,42 @@
...
+        description: 'Air leaks',
+        outcome: 'P',
+      },
+      {
+        code: 'S4',
+        description: 'Mudguards condition',
+      },
+    ];
+  }
+}
```
rossfauldsbjss: Doesn't need a singleton for this, or indeed a class.  Just export the function getMalformedVehicleChecks and import and call that in the spec file.

faggc (approved):

rossfauldsbjss (approved):
looks good.

TomBillingtonUK (approved):

PR #1327, opened 2020-02-24
MES-4118 3.2.0 Back-port for Vehicle Checks Fix

tomcrawleyy:
## Description
**Issue**
If a DE selects a ShowMeTellMe Question but does not provide an outcome for the question before terminating, the question will show up as a fault on debrief page.
**Fix**
NgOnInit added to VehicleChecksCard to capture and remove any ShowMeTellMe provided without an outcome from the state.
**Explanation**
This in turn will mean that the vehicle checks card will not show if a DE selected a question and did not provide an outcome for that question before terminating the test. Previously; it would include the question as a fault if the outcome didn't match to a pass, therefore if the property did not exist it would mark it as a fault, meaning the vehicle checks card would show in instances where unnecessary/unwanted.
**Note**
If a DE selects 2 or more ShowMeTellMe Questions, and provides an outcome for only 1 of the questions, only the question with an outcome provided will display in the debrief page.
## Checklist

### DIFF
--- a/src/pages/debrief/components/vehicle-checks-card/__mocks__/vehicle-checks-card.mock.ts
+++ b/src/pages/debrief/components/vehicle-checks-card/__mocks__/vehicle-checks-card.mock.ts
@@ -1,42 +1,28 @@
 import { QuestionResult } from '@dvsa/mes-test-schema/categories/common';
 
-export class VehicleChecksCardMock {
-
-  private static instance: VehicleChecksCardMock;
-
-  private constructor() { }
-
-  public static getInstance(): VehicleChecksCardMock {
-    if (!VehicleChecksCardMock.instance) {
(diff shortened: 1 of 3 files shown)


src/pages/debrief/components/vehicle-checks-card/__mocks__/vehicle-checks-card.mock.ts
```diff
@@ -0,0 +1,42 @@
...
+        description: 'Air leaks',
+        outcome: 'P',
+      },
+      {
+        code: 'S4',
+        description: 'Mudguards condition',
+      },
+    ];
+  }
+}
```
rossfauldsbjss: Doesn't need a singleton for this, or indeed a class. Just export the function getMalformedVehicleChecks and import and call that in the spec file.

ammarhaiderbjss (approved):

PR #1340, opened 2020-02-26
Fixed transmission banner not showing on pass finalisation, and pre populated value

edwardwilson:
## Description
Fixes the transmission banner not showing on first selection in Mod 1
MES-5003 

## Checklist

### DIFF
--- a/src/pages/pass-finalisation/cat-a-mod1/pass-finalisation.cat-a-mod1.page.html
+++ b/src/pages/pass-finalisation/cat-a-mod1/pass-finalisation.cat-a-mod1.page.html
@@ -24,7 +24,6 @@
 
         <transmission
           [formGroup]="form"
-          [transmission]="pageState.transmission$ | async"
           (transmissionChange)="transmissionChanged($event)">
         </transmission>
 


src/pages/pass-finalisation/cat-a-mod1/pass-finalisation.cat-a-mod1.page.html
```diff
@@ -24,7 +24,6 @@
 
         <transmission
           [formGroup]="form"
-          [transmission]="pageState.transmission$ | async"
           (transmissionChange)="transmissionChanged($event)">
         </transmission>
```
rossfauldsbjss, line 29: This feels like a bit of a hack. There must be a better way than deliberately not passing in state.

edwardwilson: I think this is the most appropiate way, it's a one directional binding

rossfauldsbjss: So - from the description there is no transmission for a cat-a-mod1.  So why is there a transmission component on the page.  If there is no transmission and yet the transmission component is responsible for displaying a banner this definitely doesn't seem right.  The banner display should be separated out from the component and displayed conditionally on the page.

rossfauldsbjss: Removed objection - old title made me think there was no transmission on this test.

rossfauldsbjss (approved):
looks good

appleabs (approved):

PR #1346, opened 2020-02-27
Fix uppercase directive

Cullima:
## Description
- Simplifies InputRestrictionUppercaseAlphanumDirective directive and removes deprecated [initEvent](https://developer.mozilla.org/en-US/docs/Web/API/Event/initEvent) method.
- Fixes error in Chrome where completing Vehicle Registration with Alpha chars only fails to update NGRX store 

## Checklist

### DIFF
--- a/src/directives/__tests__/input-restriction-uppercasealphanum.directive.spec.ts
+++ b/src/directives/__tests__/input-restriction-uppercasealphanum.directive.spec.ts
@@ -22,7 +22,7 @@ describe('Directive: InputRestrictionUppercaseAlphanumDirective', () => {
     });
     fixture = TestBed.createComponent(TestAlphaNumComponent);
   });
-  it('should remove spaces, non alpha numeric chars and upper case input', () => {
+  it('should upper case input and remove spaces and non alpha numeric chars', () => {
     fixture.detectChanges();
     inputElement = fixture.debugElement.query(By.css('input'));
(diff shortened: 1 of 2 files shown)


szabikr (approved):

src/directives/__tests__/input-restriction-uppercasealphanum.directive.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      declarations: [
+        TestAlphaNumComponent,
+        InputRestrictionUppercaseAlphanumDirective,
+      ],
+    });
+    fixture = TestBed.createComponent(TestAlphaNumComponent);
+  });
+  it('should remove spaces, non alpha numeric chars and upper case input', () => {
```
rossfauldsbjss: test description made me think it would remove upper case - possibly change to 'should upper lowercase letters and remove spaces, and non alpha numeric chars' (or something like that)

rossfauldsbjss (approved):
looks good.

src/directives/__tests__/input-restriction-uppercasealphanum.directive.spec.ts
Cullima: updated test description, thanks @rossfauldsbjss

PR #1400, opened 2020-03-05
Mes 4900 state management setup

mgoldson:
## Description
Adds home categories(F,G,H,K) to the Reducer Factory, including reducer composition of:

 - Tests
 - Test Data
 - Journal Data
 - Vehicle Details

state for Vehicle Checks and Highway Safety Code Reducers currently return null and reducers will be completed in specific stories associated with that functionality.

## Checklist

### DIFF
--- a/src/modules/tests/test-data/cat-home/controlled-stop/__tests__/controlled-stop.reducer.spec.ts
+++ b/src/modules/tests/test-data/cat-home/controlled-stop/__tests__/controlled-stop.reducer.spec.ts
@@ -1,0 +1,74 @@
+import { CatFUniqueTypes } from '@dvsa/mes-test-schema/categories/F';
+import { controlledStopReducer } from '../controlled-stop.reducer';
+import {
+  ToggleControlledStop,
+  ControlledStopAddDrivingFault,
+  ControlledStopAddSeriousFault,
+  ControlledStopAddDangerousFault,
+  ControlledStopRemoveFault,
+  AddControlledStopComment,
+} from '../controlled-stop.actions';
+import { CompetencyOutcome } from '../../../../../../shared/models/competency-outcome';
+
+describe('Controlled Stop Reducer' , () => {
+
+  describe('TOGGLE_CONTROLLED_STOP', () => {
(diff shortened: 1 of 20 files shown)


src/modules/tests/test-data/cat-home/controlled-stop/__tests__/controlled-stop.reducer.spec.ts
```diff
@@ -0,0 +1,74 @@
+import { CatFUniqueTypes } from '@dvsa/mes-test-schema/categories/F';
```
TomBillingtonUK, line 1: Can we not have a common controlled stop reducer for B and Home , objects are the same so it should be easy to achive.

TomBillingtonUK: Happy for this to be chucked into the tech debt list - to get this in though

src/modules/tests/vehicle-details/cat-home/vehicle-details.cat-home.reducer.ts
```diff
@@ -0,0 +1,34 @@
+import * as vehicleDetailsActions from '../common/vehicle-details.actions';
+import { VehicleDetails } from '@dvsa/mes-test-schema/categories/common';
+import { createFeatureSelector } from '@ngrx/store';
```
TomBillingtonUK: this should be using a common one as it's a common type

TomBillingtonUK (approved):
Overall looks good, I think there is some duplication which can be removed , it can be in this PR or a second follow up depending on how urgently we need this.

src/modules/tests/test-data/cat-home/controlled-stop/__tests__/controlled-stop.reducer.spec.ts
mgoldson: Yeah, good shout. I will do it as a separate PR though as we would making a change to category in prod. Better traceability that way.

faggc (approved):

PR #1447, opened 2020-03-11
Housekeeping/vehicle check constants

edwardwilson:
## Description
Small PR to clean up the vehicle check constants making clear why we disable the max-length lint rule and keeping the files consistant between categories.


faggc (approved):

rossfauldsbjss (approved):
looks good - approving on the basis that non automated testing has been conducted to ensure no impact on the categories updated.

PR #1491, opened 2020-03-23
MES-4238 - Stopped Toggle Analytics when it is programatically fired

TomBillingtonUK:
## Description
Pam noticed when we fire a TOGGLE_SERIOUS_MODE , TOGGLE_DANGEROUS_MODE and TOGGLE_REMOVE_FAULT_MODE action we always send something to GA no matter if the event is user-generated or programmatically generated.  Which is incorrect

This PR adds a toggle which allows us to decide if the action should trigger analytics or not.

I also fixed the Analytic Not Fired action as it had the same title as the Analytic Fired Action

## Checklist

### DIFF
--- a/src/pages/test-report/__tests__/test-report.analytics.effects.spec.ts
+++ b/src/pages/test-report/__tests__/test-report.analytics.effects.spec.ts
@@ -40,7 +40,7 @@ import {
   manoeuvreCompetencyLabels,
   manoeuvreTypeLabels,
 } from '../../../shared/constants/competencies/catb-manoeuvres';
-import { AnalyticRecorded } from '../../../providers/analytics/analytics.actions';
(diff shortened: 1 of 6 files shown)


szabikr (approved):
Good spot! The only thing with this is that we are not going to be able to reproduce what happened in the app from the Analytics events as we don't save everything.
I approve if that is what the expected behaviour should be

TomBillingtonUK:
From chatting with Pam you can still replicate. Previously the event stream would be:

Toggle Serious Mode
Record Serious Fault
Toggle Serious Mode (this is something the user doesn't do)
Record Driving Fault

If you followed that exactly you would get stuck in Serious Mode. So I think it makes it easier to follow as that second Toggle Serious Mode action won't be there

rossfauldsbjss (approved):
looks good

PR #1504, opened 2020-03-26
MES-4990 - Detailed Test View of Search Results

tomcrawleyy:
## Description
Refined the Detailed Test View page to fit with Cat Home tests.
## Checklist

## Screenshots (optional)
Screenshot 2020-03-26 at 09 54 08
Screenshot 2020-03-26 at 09 54 27
Screenshot 2020-03-26 at 09 54 49
Screenshot 2020-03-26 at 09 55 10
Screenshot 2020-03-26 at 09 55 29


mgoldson:
Could you add a screenshot please?

mgoldson:
Do we really need to create new components for the likes of vehicle-details-card? There is a components directory in view-test-result which you should be able to pull those common components from.

tomcrawleyy:
> Do we really need to create new components for the likes of vehicle-details-card? There is a components directory in view-test-result which you should be able to pull those common components from.

The common vehicle-details-card takes in a Cat BE Vehicle Details, although for home tests, the getVehicleLength and getVehicleWidth values don't exist and using the common vehicle-details-card (without modification) would result in those unnecessary fields showing up unpopulated in the html, i suppose this could be handled through a switch statement, but i figured since it had its own debrief card to handle the category specific changes that this one warranted its own as well

tomcrawleyy:
> Could you add a screenshot please?

screenshots added

szabikr (dismissed):

mgoldson:
> > Do we really need to create new components for the likes of vehicle-details-card? There is a components directory in view-test-result which you should be able to pull those common components from.
> 
> The common vehicle-details-card takes in a Cat BE Vehicle Details, although for home tests, the getVehicleLength and getVehicleWidth values don't exist and using the common vehicle-details-card (without modification) would result in those unnecessary fields showing up unpopulated in the html, i suppose this could be handled through a switch statement, but i figured since it had its own debrief card to handle the category specific changes that this one warranted its own as well

I would try and use the existing components if possible (appreciate they might need modifying) but your call.

mgoldson (dismissed):

src/pages/view-test-result/components/vehicle-details-card/vehicle-details-card.ts
```diff
@@ -31,6 +35,18 @@ export class VehicleDetailsCardComponent {
     );
   }
 
+  public shouldHideDimensions() : boolean {
+    switch (this.category) {
+      case TestCategory.F:
+      case TestCategory.G:
+      case TestCategory.H:
+      case TestCategory.K:
```
mgoldson, line 43: Don't the ADI categories need to be included in this? If so, we should add a test in the vehicle details card spec too.

mgoldson (approved):
Good work! See question about the ADI categories.

src/pages/view-test-result/components/vehicle-details-card/vehicle-details-card.ts
tomcrawleyy: There is a ticket duplicated from this one in the first CI sprint which covers ADI 2 changes to the detailed test view, my assumption is that it will be addressed there.

andrewsetterfield (approved):

PR #1613, opened 2020-06-18
MES - 5249 Search results driver type label

andrewsetterfield:
# Description

added function to search results to swap out Driver for Rider label for Cat A tests

## Checklist

### DIFF
--- a/src/pages/test-results-search/components/search-result/__tests__/search-result.spec.ts
+++ b/src/pages/test-results-search/components/search-result/__tests__/search-result.spec.ts
@@ -38,45 +38,25 @@ describe('SearchResultComponent', () => {
   });
 
   describe('driverRider', () => {
+    const searchResult = {
+      costCode: null,
+      testDate: null,
+      driverNumber: null,
+      candidateName: null,
+      applicationReference: null,
+      category: null,
+      activityCode: null,
+    };
     it('should return driver when cat is not A', () => {
-
-      const searchResult = {
-        costCode: null,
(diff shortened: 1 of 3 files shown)


src/pages/test-results-search/components/search-result/__tests__/search-result.spec.ts
```diff
@@ -36,4 +37,48 @@ describe('SearchResultComponent', () => {
...
+        driverNumber: null,
+        candidateName: null,
+        applicationReference: null,
+        category: TestCategory.EUAM2,
+        activityCode: null,
+      };
+      expect(component.driverRider(searchResult)).toEqual('Rider');
+    });
+  });
+
```
tomcrawleyy: this search result constant could get moved under the describe so that its only instantiated once, then you could just update the searchResult.category at the start of each it function, would make for a cleaner commit

src/pages/test-results-search/components/search-result/search-result.ts
```diff
@@ -41,6 +42,27 @@ export class SearchResultComponent {
     return `${name.title} ${name.firstName} ${name.lastName}`;
   }
 
+  driverRider(searchResult: SearchResultTestSchema): string {
```
Cullima, line 45: is there not already a helper for this, which you can just pass the category code to? 
src/shared/helpers/driver-type.ts

Cullima (changes requested):

PR #1692, opened 2020-07-27
Mes-5339: Delegated activity codes

andrewsetterfield:
## Description

Add new activity codes if delegated examiner test for some categories.
https://jira.dvsacloud.uk/browse/MES-5339

## Checklist


src/pages/non-pass-finalisation/cat-a-mod1/non-pass-finalisation.cat-a-mod1.page.ts
```diff
@@ -91,7 +91,7 @@ export class NonPassFinalisationCatAMod1Page extends BasePageComponent implement
   ) {
     super(platform, navController, authenticationProvider);
     this.form = new FormGroup({});
-    this.activityCodeOptions = activityCodeModelList;
```
matthew2564, line 94: is there a need for changing this for a category a DLG examiner won't complete? same for B/MOD2/ADI

package.json
```diff
@@ -224,4 +224,4 @@
       "npm run lint:fix"
     ]
   }
-}
+}
```
andrewsetterfield: need to correct this

src/pages/non-pass-finalisation/cat-a-mod1/non-pass-finalisation.cat-a-mod1.page.ts
andrewsetterfield: yes, yes it it

matthew2564: so you cant just use what was there before? i.e.

this.activityCodeOptions = activityCodeModelList;

andrewsetterfield: actually no.....it doesnt need to be the same. if i leave it as before it will default to true and add the additional values. So as such i need to set it to false.

andrewsetterfield: same for all none required categories, ie B, HOME.

matthew2564: are you not able to change the default to no then?

andrewsetterfield: no, cant change the default to false.  If i do then i need to change every office page and the test-selector and if i change the test selector then im back where i was friday afternoon with the value no patching correctly.  This may be solved with having two completely seperate functions within activity-codes constants and two different selectors.

Cullima (approved):

tomcrawleyy (approved):

PR #1828, opened 2020-11-04
mes-6053: Health Declaration bug

andrewsetterfield:
## Description

updated health declration to always be a boolean not a string containing true or false
https://jira.dvsacloud.uk/browse/MES-6053

## Checklist

### DIFF
--- a/src/components/common/health-declaration-signed/health-declaration-signed.html
+++ b/src/components/common/health-declaration-signed/health-declaration-signed.html
@@ -15,7 +15,7 @@
                  class="gds-radio-button"
                  [class.ng-invalid]="invalid"
                  (change)="healthDeclarationChanged($event.target.value)"
-                 value="true">
+                 value="Signed">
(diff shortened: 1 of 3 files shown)


ammarhaiderbjss (approved):
Nice use of an enum there!

src/components/common/health-declaration-signed/health-declaration-signed.html
```diff
@@ -15,7 +15,7 @@
                  class="gds-radio-button"
                  [class.ng-invalid]="invalid"
                  (change)="healthDeclarationChanged($event.target.value)"
-                 value="true">
+                 value="Signed">
```
matthew2564, line 18: could we not have just done [value]="true" which i think should leave it as a boolean value? this would then eliminate the need for the enum and the elvis logic in the TS

andrewsetterfield: unfortunatly that didnt seem to work :(

matthew2564 (approved):

PR #1844, opened 2020-11-16
MES-5910 Journal Cache

andrewsetterfield:
## Description
Introduces logic to prevent the user from starting tests they have already done when the cache is empty and the journal is retrieved.

## Checklist


src/pages/journal/journal.ts
```diff
@@ -171,6 +178,64 @@ export class JournalPage extends BasePageComponent implements OnInit {
...
+   * @param emission
+   */
+  getSearchResults = (emission: SlotItem[]) => {
+    const advancedSearchParams: AdvancedSearchParams = {
+      startDate: '',
+      endDate: '',
+      staffNumber: removeLeadingZeros('1234567'),
+      costCode: '',
+    };
+    forkJoin([
```
ammarhaiderbjss: do we need `forkJoin`?

src/pages/journal/journal.ts
```diff
@@ -171,6 +178,64 @@ export class JournalPage extends BasePageComponent implements OnInit {
...
+      applicationId: slot.slotData.booking.application.applicationId,
+      bookingSequence: slot.slotData.booking.application.bookingSequence,
+      checkDigit: slot.slotData.booking.application.checkDigit,
+    };
+
+    return references.some((reference) => {
+      return reference === parseInt(formatApplicationReference(applicationReference), 10);
+    });
+  }
+
```
ammarhaiderbjss: Perhaps you were already planning on this, but I think the work to retrieve the app refs should be in a separate provider to keep things clean. The API for the provider could be -

- Takes a startDate, endDate and staffNumber
- Returns a list of completed app refs

The journal page would only then be responsible for the business logic of matching completed app refs to the journal bookings.

andrewsetterfield: fully agree. will look to refactor.

src/pages/journal/journal.ts
andrewsetterfield: no longer required, will remove.

src/pages/journal/journal.ts
```diff
@@ -171,6 +178,64 @@ export class JournalPage extends BasePageComponent implements OnInit {
...
+    const applicationReference: ApplicationReference = {
+      applicationId: slot.slotData.booking.application.applicationId,
+      bookingSequence: slot.slotData.booking.application.bookingSequence,
+      checkDigit: slot.slotData.booking.application.checkDigit,
+    };
+
+    return references.some((reference) => {
+      return reference === parseInt(formatApplicationReference(applicationReference), 10);
+    });
+  }
```
ammarhaiderbjss: Another idea for this part of the solution would be to simply update the test statuses themselves. You can use the existing `SetTestStatusSubmitted` action to set the status of the submitted tests to `Submitted`.

Then if you call the `PersistTests` action this will save the new statuses to the cache for you.
This way the journal should naturally try update its view to display the activity code of the test rather than the "Start test" button. 

There might be some typing issues with this approach though.

src/pages/journal/journal.ts
```diff
@@ -67,6 +78,9 @@ export class JournalPage extends BasePageComponent implements OnInit {
   start = '2018-12-10T08:10:00+00:00';
   merged$: Observable<void | number>;
   todaysDate: DateTime;
+  searchResults: SearchResultTestSchema[] = [];
+  searchResultsAppRefs: number[];
+  cachedTests: boolean;
```
ammarhaiderbjss: This prop could have a better name, perhaps `isCachedTests` as it is a boolean?

src/modules/tests/tests.model.ts
```diff
@@ -9,4 +9,5 @@ export interface TestsModel {
   currentTest: CurrentTest;
   startedTests: { [slotId: string]: TestResultSchemasUnion };
   testStatus: { [slotId: string]: TestStatus };
+  completedTests: number[];
```
szabikr: a number?

src/pages/journal/journal.ts
```diff
@@ -171,6 +207,70 @@ export class JournalPage extends BasePageComponent implements OnInit {
...
+   */
+  getSearchResults = (emission: SlotItem[]) => {
+    const advancedSearchParams: AdvancedSearchParams = {
+      startDate: moment().subtract(14, 'days').format('YYYY-MM-DD'),
+      endDate: moment().format('YYYY-MM-DD'),
+      staffNumber: removeLeadingZeros(this.staffNumber),
+      costCode: '',
+    };
+    return new Promise((resolve) => {
+      this.searchProvider.advancedSearch(advancedSearchParams)
```
szabikr: This should happen in the Journal Effect. Don't think that making the Journal Page dependent on SearchProvider is a good idea.

szabikr: As a matter of fact it should be in its own effect. Let's keep to the Open-closed principle

szabikr (changes requested):
Baking this cache logic into the Journal page is not a great idea, makes existing code more difficult to understand and unpredictable. Should be looking at extending existing components rather than modifying them.

And we were talking about having this logic in the core of other features when rolling out the new app for example, so we need a bit more robust solution for this.

I'm happy to take a look and flash out an alternative with you @andrewsetterfield and @tomcrawleyy

src/modules/tests/tests.actions.ts
```diff
@@ -102,6 +104,11 @@ export class SendCurrentTestFailure implements Action {
   constructor(public isDuplicateUpload: boolean) { }
 }
 
+export class AddCompletedTests implements Action {
+  readonly type = ADD_COMPLETED_TESTS;
+  constructor(public references: number[]) {}
```
ammarhaiderbjss: are these "references" the application references?

src/pages/journal/journal.ts
```diff
@@ -171,6 +207,70 @@ export class JournalPage extends BasePageComponent implements OnInit {
...
+
+  /**
+   * Call endpoint to obtain users previously completed tests and
+   * pass to createSlots function for evaluation
+   * @param emission
+   */
+  getSearchResults = (emission: SlotItem[]) => {
+    const advancedSearchParams: AdvancedSearchParams = {
+      startDate: moment().subtract(14, 'days').format('YYYY-MM-DD'),
+      endDate: moment().format('YYYY-MM-DD'),
```
ammarhaiderbjss: We should read the values from config to create this time window (`journal.numberOfDaysToView`) as they can feasibly change in the future.

src/pages/journal/journal.ts
szabikr: forkJoin is not present anymore

src/pages/journal/journal.ts
szabikr: This functionality has been moved to the journal effects now

src/pages/journal/journal.ts
szabikr: Haven't approached it this way, but happy to discuss this further if the solution that is now implemented seems significantly weaker than what you described.

src/pages/journal/journal.ts
szabikr: `isCachedTests` is no longer present

src/modules/tests/tests.model.ts
szabikr: Changed it to `SearchResultsTestSchema`

src/pages/journal/journal.ts
szabikr: Functionality has been moved to `journal.effects`

src/modules/tests/tests.actions.ts
szabikr: `AddCompletedTests` is no longer present

src/modules/journal/journal.actions.ts
```diff
@@ -10,6 +11,10 @@ export const LOAD_JOURNAL_FAILURE = '[JournalEffects] Load Journal Failure';
 export const LOAD_JOURNAL_SILENT = '[JournalEffect] Load Journal Silent';
 export const LOAD_JOURNAL_SILENT_FAILURE = '[JournalEffect] Load Journal Silent Failure';
 
+export const LOAD_COMPLETED_TESTS = '[JoarnalEffect] Load Completed Tests';
+export const LOAD_COMPLETED_TESTS_SUCCESS = '[JoarnalEffect] Load Completed Tests Success';
+export const LOAD_COMPLETED_TESTS_FAILURE = '[JoarnalEffect] Load Completed Tests Failure';
```
tomcrawleyy: small spelling error, should be [JournalEffect] not [JoarnalEffect]

src/pages/journal/journal.ts
szabikr: Is `journal.numberOfDaysToView` going to return the value representing navigating forwards and backwards combined? Or just one way?

tomcrawleyy (approved):

piercemorris (approved):

src/modules/journal/journal.actions.ts
szabikr: corrected the spelling mistake

szabikr (approved):
I made made most of the requested changes however, I think it's good now

src/pages/journal/journal.ts
szabikr: Getting the number of days from config

PR #1853, opened 2020-11-27
MES-6084 remove safety questions

szabikr:
## Description

Removing SafetyQuestions from cat d vehicle checks, office and more.

There is an additional change that I had to introduce to AMod2 Office page where by the component was using `SafetyQuestionResult` from the common types which has been removed from the `@dvsa/mes-test-schemas` (shouldn't have been there at the first place). So, I changed it to use the `QuestionResult` from the AMod2 schemas which is the correct type.

## Checklist


piercemorris (approved):

ammarhaiderbjss (approved):
So satisfying to remove code :)

PR #2029, opened 2021-10-25
MES-7290 / Fault count summary

matthew2564:
## Description
- Additional AC's were added to 7290 to update existing logic for fault count and fault summary. Previously, these had been driven via the category, although now with the new logic of being able to switch between the number of SM/TM questions on the extended tests, the logic is now driven from a state variable

## Checklist

### DIFF
--- a/src/pages/waiting-room-to-car/cat-d/components/vehicle-checks-modal/vehicle-checks-modal.cat-d.page.ts
+++ b/src/pages/waiting-room-to-car/cat-d/components/vehicle-checks-modal/vehicle-checks-modal.cat-d.page.ts
@@ -59,6 +59,7 @@ import {
   getVehicleChecksCatD,
   getSelectedShowMeQuestions,
   getSelectedTellMeQuestions,
+  CatDVehicleChecks,
 } from '../../../../../modules/tests/test-data/cat-d/vehicle-checks/vehicle-checks.cat-d.selector';
 import {
   getSafetyQuestionsCatD,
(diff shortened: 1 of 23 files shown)


src/pages/waiting-room-to-car/cat-d/components/vehicle-checks-modal/vehicle-checks-modal.cat-d.page.ts
```diff
@@ -171,23 +157,69 @@ export class VehicleChecksCatDModal {
...
+        break;
+      case TestCategory.DE:
+      case TestCategory.D1E:
+        numberOfShowMeQuestions = this.getNumberOfShowMeQuestions(fullLicenceHeld);
+        numberOfTellMeQuestions = this.getNumberOfTellMeQuestions(fullLicenceHeld);
+        break;
+    }
+    this.showMeQuestionsNumberArray = Array(numberOfShowMeQuestions);
+    this.tellMeQuestionsNumberArray = Array(numberOfTellMeQuestions);
   }
```
solomonsimukoko, line 223: Is the a way we can make this function common so that for the CAT C can just have its own case

solomonsimukoko (approved):

andrewsetterfield (approved):

src/pages/waiting-room-to-car/cat-d/components/vehicle-checks-modal/vehicle-checks-modal.cat-d.page.ts
matthew2564: That is an improvement we could look to make in DES4, but think we should aim to make as little change as possible here

PR #2050, opened 2021-11-24
MES-7424 | Display full licence held on search completed test results

gittins7:
## Description
Added conditional field to test-details-card to display fullLicenceHeld
## Checklist

## Screenshots (optional)
Screenshot 2021-11-24 at 15 40 27

### DIFF
--- a/src/pages/view-test-result/cat-c/view-test-result.cat-c.page.ts
+++ b/src/pages/view-test-result/cat-c/view-test-result.cat-c.page.ts
@@ -132,6 +132,7 @@ export class ViewTestResultCatCPage extends BasePageComponent implements OnInit 
       entitlementCheck: this.testResult.journalData.testSlotAttributes.entitlementCheck,
       slotType: this.testResult.journalData.testSlotAttributes.slotType,
       previousCancellations: get(this.testResult, 'journalData.testSlotAttributes.previousCancellation', []),
+      fullLicenceHeld: this.testResult.testData.vehicleChecks.fullLicenceHeld,
     };
   }
 
--- a/src/pages/view-test-result/cat-c/view-test-result.cat-c.page.ts
+++ b/src/pages/view-test-result/cat-c/view-test-result.cat-c.page.ts
@@ -132,7 +132,7 @@ export class ViewTestResultCatCPage extends BasePageComponent implements OnInit 
       entitlementCheck: this.testResult.journalData.testSlotAttributes.entitlementCheck,
       slotType: this.testResult.journalData.testSlotAttributes.slotType,
       previousCancellations: get(this.testResult, 'journalData.testSlotAttributes.previousCancellation', []),
-      fullLicenceHeld: this.testResult.testData.vehicleChecks.fullLicenceHeld,
+      fullLicenceHeld: get(this.testResult, 'testData.vehicleChecks.fullLicenceHeld'),
     };
   }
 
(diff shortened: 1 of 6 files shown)


src/pages/view-test-result/cat-c/view-test-result.cat-c.page.ts
```diff
@@ -132,6 +132,7 @@ export class ViewTestResultCatCPage extends BasePageComponent implements OnInit
       entitlementCheck: this.testResult.journalData.testSlotAttributes.entitlementCheck,
       slotType: this.testResult.journalData.testSlotAttributes.slotType,
       previousCancellations: get(this.testResult, 'journalData.testSlotAttributes.previousCancellation', []),
+      fullLicenceHeld: this.testResult.testData.vehicleChecks.fullLicenceHeld,
```
matthew2564: I would wrap this in Lodash get to be safe defaulting to false i.e.

```fullLicenceHeld: get(this.testResult, 'testData.vehicleChecks.fullLicenceHeld', false)```

File: src/pages/view-test-result/cat-d/view-test-result.cat-d.page.ts [hunk 2]
```diff
@@ -132,6 +132,7 @@ export class ViewTestResultCatDPage extends BasePageComponent implements OnInit
       entitlementCheck: this.testResult.journalData.testSlotAttributes.entitlementCheck,
       slotType: this.testResult.journalData.testSlotAttributes.slotType,
       previousCancellations: get(this.testResult, 'journalData.testSlotAttributes.previousCancellation', []),
+      fullLicenceHeld: this.testResult.testData.vehicleChecks.fullLicenceHeld,
```
matthew2564: same as above

src/pages/view-test-result/components/test-details-card/test-details-card.ts
```diff
@@ -16,4 +16,8 @@ export class TestDetailsCardComponent {
     return specialNeedsArray.length > 0 && specialNeedsArray[0] !== 'None';
   }
 
+  showFullCatHeld(): boolean {
+    return this.data.category.includes('+E');
```
matthew2564: This is a common component so we need to be more specific here. This would currently display the field for B+E which we do not want. Should do something like

```return this.data.category === TestCategory.CE || this.data.category === TestCategory.C1E || this.data.category === TestCategory.DE || this.data.category === TestCategory.D1E;```

Or if got a reservation about the OR's, then something like

```return [TestCategory.CE, TestCategory.C1E, TestCategory.DE, TestCategory.D1E].some((category) => this.data.category);```

Comment (andrewsetterfield) [reply src/pages/view-test-result/components/test-details-card/test-details-card.ts]: might be worth making a helper function to identify these categories as its likely to come up again

File: src/pages/view-test-result/components/test-details-card/test-details-card.html [hunk 4]
```diff
@@ -8,6 +8,7 @@ <h4>Test details</h4>
       <data-row [label]="'Time'" [value]="data.time"></data-row>
       <data-row [label]="'Application reference'" [value]="data.applicationReference"></data-row>
       <data-row [label]="'Test category'" [value]="data.category"></data-row>
+      <data-row *ngIf="showFullCatHeld()" [label]="'Full Cat [cat-E] held?'" [value]="data?.fullLicenceHeld ? 'Yes' : 'No'"></data-row>
```
matthew2564: ```[cat-E]``` needs to be dynamic based on category

```[label]="'Full cat ' + data.category | slice: 0: 1 + ' held?'"```

src/pages/view-test-result/components/test-details-card/test-details-card.ts
matthew2564: @andrewsetterfield There are helpers in the doc service so this is something we could introduce into the FE. But lets leave that for another ticket .

FYI - https://github.com/dvsa/mes-documents-service/blob/develop/src/functions/sendCandidateResults/application/service/template-id-provider.ts

gittins7: was getting a typing error when using .some() so for convenience i've done it with the inverse use of .includes() instead.

Tested it and works as required but let me know if there's any reasons not to use this and to stick with using .some()

matthew2564 (approved):

andrewsetterfield (approved):

